PR #9, opened 2023-03-04
FE: Add basic collaboration

phlmn:


anuejn (approved):
for me it is okay to merge this as-is.
though we decided to switch away from Y.js it might still be of use to have this code in main for future reference

PR #5, opened 2023-03-03
initial worker skeleton

anuejn:

PR #13, opened 2023-03-04
fix transcribee-proto package

rroohhh:


proto/transcribee_proto/document.py
```diff
@@ -3,7 +3,7 @@
 from pydantic import BaseModel
 
 
-class TranscriptSegment(BaseModel):
+class Token(BaseModel):
```
anuejn, line 6: I much prefer the TranscriptSegment name because these dont nescessarily have to correspond to tokens once transcript correction happened

proto/pyproject.toml
```diff
@@ -16,7 +16,9 @@ requires-python = ">=3.10"
 readme = "README.md"
 license = {text = "AGPL-3.0"}
 
-[tool.pdm]
+[tool.pdm.build]
+package-dir = "./"
+includes = ["transcribee_proto"]
```
anuejn, line 21: ah, nice, thanks for figuring that out :)

anuejn (changes requested):

proto/transcribee_proto/document.py
anuejn: maybe something like shrapnel, part, atom, bit slice, piece, hunk?

rroohhh: Yeah, `Atom` works for me.

rroohhh:
folded into #5

PR #19, opened 2023-03-09
Sync Protocol & Small Editor Improvements

pajowu:
This PR adds a simple syncing protocol. For now, the main advantage is that the backend can signal once the backlog of sync messages is fully send to the client, so the client can indicate that it is still catching up to the user (currently done by blurring the editor).

This protocol can be easily extended later for the full sync protocol of our CRDT (incremental updates, combining multiple updates into one message, ...)


frontend/src/editor/TranscriptionEditor.tsx
```diff
@@ -73,8 +75,20 @@ export default function TranscriptionEditor() {
...
+    const preventCtrlS = (e: KeyboardEvent) => {
+      const ctrlOrCmd = window.navigator.platform.match("Mac") ? e.metaKey : e.ctrlKey
+      if (ctrlOrCmd && e.key === 's') {
+        e.preventDefault();
+        console.log('CommandOrControl + S prevented – we automatically save the document anyways');
+      }
+    };
+    document.addEventListener('keydown', preventCtrlS);
+    return () => document.removeEventListener('keydown', preventCtrlS)
+  })
```
rroohhh, line 88: This should have a dependency key or it will get executed every time, right?

pajowu: oops, good catch :D

frontend/src/editor/TranscriptionEditor.tsx
```diff
@@ -73,8 +75,20 @@ export default function TranscriptionEditor() {
...
+        console.log('CommandOrControl + S prevented – we automatically save the document anyways');
+      }
+    };
+    document.addEventListener('keydown', preventCtrlS);
+    return () => document.removeEventListener('keydown', preventCtrlS)
+  }, [])
+
   return (
-    <div>
+    <div className={syncComplete ?'':'blur'}>
```
phlmn, line 91: Looks to me as if this was not formattted with prettier. I also noticed we are not running prettier on pre-commit. Just created a PR for this: #31

phlmn (commented):
Needs a `pnpm run format`. Otherwise LGTM!

phlmn (approved):

PR #34, opened 2023-03-10
🔧 Put token auth first

pajowu:
Otherwise we get CRSF errors with a user is both logged in via session auth and provides a token


phlmn (approved):
Nice!

PR #35, opened 2023-03-10
Connect Worker to Backend

pajowu:


proto/transcribee_proto/api.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TaskBase(BaseModel):
+    id: str
+    document: Document
+    task_type: TaskType
+
+
+class DiarizeTask(TaskBase):
+    task_type: Literal[TaskType.DIARIZE]
```
anuejn, line 30: there should be a `= TaskType.DIARIZE`

proto/transcribee_proto/api.py
```diff
@@ -0,0 +1,49 @@
...
+    task_parameters: Dict[str, Any]
+
+
+class TranscribeTaskParameters(BaseModel):
+    lang: str
+    model: str
+
+
+class TranscribeTask(TaskBase):
+    task_type: Literal[TaskType.TRANSCRIBE]
```
anuejn, line 40: see above

proto/transcribee_proto/api.py
```diff
@@ -0,0 +1,49 @@
...
+    task_parameters: TranscribeTaskParameters
+
+
+class AlignTask(TaskBase):
+    task_type: Literal[TaskType.ALIGN]
```
anuejn, line 45: see above

worker/run.py
```diff
@@ -0,0 +1,32 @@
...
+import logging
+
+from transcribee_proto.api import TaskType
+from transcribee_worker.worker import Worker
+
+logging.basicConfig(level=logging.INFO)
+
+
+async def main():
+    parser = argparse.ArgumentParser(description="Worker")
```
anuejn, line 14:
```suggestion
    parser = argparse.ArgumentParser(description="the worker doing ML inference for the transcribee open source transcription platform")
```

worker/transcribee_worker/worker.py
```diff
@@ -0,0 +1,152 @@
...
+
+import requests
+from pydantic import parse_raw_as
+from transcribee_proto.api import AlignTask, AssignedTask, DiarizeTask
+from transcribee_proto.api import Document as ApiDocument
+from transcribee_proto.api import TaskType, TranscribeTask
+from transcribee_worker.util import load_audio
+from transcribee_worker.whisper_transcribe import transcribe
+
+logging.basicConfig(level=logging.INFO)
```
anuejn, line 16: this should be in the run.py

worker/run.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+    parser.add_argument("--token", help="Worker token", required=True)
+    args = parser.parse_args()
+
+    worker = Worker(
+        base_url=f"{args.coordinator}/api/v1/tasks",
+        token=args.token,
+        task_types=[TaskType.TRANSCRIBE],
+    )
+    await worker.run_task()
```
anuejn, line 28: can I have a while true?
also can I have sleeping if no task was executed?

anuejn (changes requested):

worker/run.py
pajowu: I changed it to `"The worker for the transcribee open source transcription platform"`, since we not (necessarily) only do ml inference in the worker

worker/transcribee_worker/whisper_transcribe.py
```diff
@@ -123,39 +124,49 @@ def handle_new_segment(
...
-    params.no_context = (
-        False  # if False, feeds back already transcribed text back to the model
+    params = (
+        api.Params.from_sampling_strategy(sampling)
+        .with_no_context(
+            False
+        )  # if False, feeds back already transcribed text back to the model
+        .with_num_threads(4)
+        .with_language(lang_code)
+        .with_max_segment_length(120)  # parameter stolen from whisper.cpp cli
```
rroohhh, line 144: Why is this changed from 60 to 120?

worker/transcribee_worker/worker.py
```diff
@@ -0,0 +1,154 @@
...
+            return
+
+        document_file = self.get_document_file(task.document)
+        audio = load_audio(str(document_file))
+
+        def progress_callback(_ctx, progress, _data):
+            self.keepalive(task.id, progress=progress / 100)
+
+        paragraphs = []
+        async for paragraph in transcribe(
```
rroohhh, line 103: This could use async list comprehension

worker/transcribee_worker/worker.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+        req.raise_for_status()
+        return parse_raw_as(Optional[AssignedTask], req.text)
+
+    def _get_tmpfile(self, filename: str) -> Path:
+        if self.tmpdir is None:
+            raise ValueError("`tmpdir` must be set")
+        return self.tmpdir / filename
+
+    def get_document_file(self, document: ApiDocument) -> Optional[Path]:
```
rroohhh, line 55: Why do we save the file to the filesystem? 
Can we not just keep it in memory?

worker/transcribee_worker/worker.py
anuejn: I belive it is fine for now as the worker throws a `NotImplemented` anyways currently.

proto/transcribee_proto/document.py
```diff
@@ -32,6 +32,12 @@ class Document(BaseModel):
     lang: str
     paragraphs: List[Paragraph]
 
+    def is_empty(self) -> bool:
```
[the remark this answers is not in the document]
anuejn: I would argue it should not contain True in this case. No matter if it is whitespace there is *something* and it is thus not empty?

worker/transcribee_worker/whisper_transcribe.py
anuejn: Can we just set this to 0? I heard this will give best results?

anuejn (approved):
I am okay with this in the current state.
If it is okay for you @pajowu and @rroohhh plz merge :)

worker/transcribee_worker/whisper_transcribe.py
rroohhh: Where did you hear this?
Why does the whisper.cpp cli set it to `60` when one specifies `0` then? (ref https://github.com/ggerganov/whisper.cpp/blob/09e90680072d8ecdf02eaf21c393218385d2c616/examples/main/main.cpp#L590)

worker/run.py
```diff
@@ -28,8 +30,17 @@ async def main():
...
-        no_work = await worker.run_task()
-        if no_work:
+        try:
+            no_work = await worker.run_task()
+            if no_work:
+                await asyncio.sleep(5)
+        except requests.exceptions.ConnectionError:
+            logging.warn("could not connect to backend")
+            await asyncio.sleep(5)
+        except Exception:
```
rroohhh, line 40: Are we sure in this case the `worker` is still valid?
Maybe we want to recreate the worker completely in this case?

worker/transcribee_worker/worker.py
```diff
@@ -52,18 +53,13 @@ def _get_tmpfile(self, filename: str) -> Path:
             raise ValueError("`tmpdir` must be set")
         return self.tmpdir / filename
 
-    def get_document_file(self, document: ApiDocument) -> Optional[Path]:
-        logging.debug(f"Getting file. {document=}")
+    def get_document_audio(self, document: ApiDocument) -> Optional[BytesIO]:
```
rroohhh, line 56: Do we want to use `Optional[T]` or `T | None`?
(I am fine either way, but a consistent style is probably nice)

(Are there linters that allow custom lint rules for stuff like this?)

rroohhh:
Apart from my comments I am fine with merging :)

rroohhh (approved):

worker/run.py
anuejn: I also thought about that (also validating the connection in the constructor) but I didn't see any immediate advantage and think it is fine like that for now

worker/transcribee_worker/worker.py
anuejn: I dont know but I think our codebase does not use the | syntax anywhere yet.
I also think that it is the only python 3.10 feature we would require.

pajowu: I'll keep optional for now, but am open to changing it at some point if we agree to use `|` in general

worker/transcribee_worker/whisper_transcribe.py
pajowu: Oops, got catch. I changed it during testing

worker/run.py
pajowu: Let's cross that bridge once we get there. Currently we also have a blanket except in `Worker.run_task` which will capture any exceptions from the actual task code, which I think would be the main source of state corruption, so this needs more work

PR #36, opened 2023-03-11
Add very basic gallery

phlmn:


backend/transcribee_backend/api/views.py
```diff
@@ -23,7 +23,7 @@
 
 class DocumentViewSet(viewsets.ModelViewSet):
     def get_queryset(self):
-        return Document.objects.filter(user=self.request.user)
+        return Document.objects.filter(user=self.request.user).order_by("-created_at")
```
pajowu, line 26: I think `-changed_at` would be the better one here, so the user always sees the document it just edited first.

pajowu: Also I think we might want to add an ordering to the document model, so we always get that ordering? `created_at` might be the right one there, so we have a more permanent sorting https://docs.djangoproject.com/en/4.1/ref/models/options/#ordering

in base/models.py:
```
class Document(models.Model):
    class Meta:
        ordering = ["-created_at", "id"]
```

frontend/index.html
```diff
@@ -4,11 +4,11 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>transcribee 🐝</title>
-    <link rel="stylesheet" href="src/index.css">
-    <link rel="stylesheet" href="node_modules/@fontsource/inter/latin.css">
+    <link rel="stylesheet" href="/src/index.css">
+    <link rel="stylesheet" href="/node_modules/@fontsource/inter/latin.css">
```
pajowu, line 8: Can we make this (and the routes) depend on a base url setting? So we can host transcribee somewhere else than the document root?

frontend/src/editor/TranscriptionEditor.tsx
```diff
@@ -47,15 +47,11 @@ export default function TranscriptionEditor() {
     console.log('new yDoc');
```
pajowu, line 47: 🖨️

frontend/src/pages/DocumentPage.tsx
```diff
@@ -1,10 +1,11 @@
+import { RouteComponentProps } from 'wouter';
 import TranscriptionEditor from '../editor/TranscriptionEditor';
 
-export default function DocumentPage() {
+export default function DocumentPage({ params }: RouteComponentProps<{ documentId: string }>) {
   return (
-    <div className="container p-8 mx-auto">
+    <div className="container p-6 mx-auto">
       <h2 className="text-2xl font-bold mb-8">Test Editor</h2>
```
pajowu, line 7: Wish (not necessarily for this PR): Show the actual document title here

frontend/src/pages/DocumentPage.tsx
```diff
@@ -1,10 +1,11 @@
+import { RouteComponentProps } from 'wouter';
 import TranscriptionEditor from '../editor/TranscriptionEditor';
 
-export default function DocumentPage() {
+export default function DocumentPage({ params }: RouteComponentProps<{ documentId: string }>) {
```
pajowu, line 4: Now that we have the gallery which serves as a "home page" I would wish for a link from the document page back to the gallery

frontend/src/pages/HomePage.tsx
```diff
@@ -1,12 +1,78 @@
...
-      <PrimaryButton onClick={() => navigate('/new')}>New Document</PrimaryButton>
+    <div className="container mx-auto p-6">
+      <div className="mb-8 flex justify-between items-center">
+        <h1 className="text-xl font-bold">transcribee</h1>
+        <div>
+          <PrimaryButton onClick={() => navigate('/login')}>Login</PrimaryButton>{' '}
+          <PrimaryButton onClick={() => navigate('/new')}>New Document</PrimaryButton>
+        </div>
+      </div>
+      <ul
```
pajowu, line 40: Not necessarily in this PR, maybe only once they get more complex: Can we split the list & list items out to separate components?

pajowu (commented):
Looks great in general, I added a few smaller change requests & wishes as inline comments.

frontend/src/pages/HomePage.tsx
```diff
@@ -37,7 +37,17 @@ export default function HomePage() {
...
+        className={clsx(
+          'grid',
+          'grid-cols-2',
+          'sm:grid-cols-3',
+          'md:grid-cols-4',
+          'lg:grid-cols-5',
+          'xl:grid-cols-6',
+          'gap-6',
+        )}
+      >
```
pajowu, line 50: What's the advantage of using clsx as opposed to a plain string here?

frontend/index.html
phlmn: Done, the router now respects vite's base path (https://vitejs.dev/guide/build.html#public-base-path).

frontend/src/pages/HomePage.tsx
phlmn: Yes, we should totally do that, when the item get's more details 👍

frontend/src/pages/HomePage.tsx
phlmn: This simply reduces the line length

phlmn:
@pajowu I would address the editor related settings when revising the editor in general. For now I would work towards a holistic protoype.

backend/transcribee_backend/sync/consumers.py
```diff
@@ -52,6 +57,7 @@ async def disconnect(self, close_code):
 
     async def receive(self, bytes_data):
         await create_update(document=self.document, update_content=bytes_data)
+        await save_doc(self.document)  # save document to update changed_at field
```
pajowu, line 60:
```suggestion
        await database_sync_to_async(self.document.save)()  # save document to update changed_at field
```

pajowu (changes requested):

pajowu (approved):

PR #38, opened 2023-03-12
🚀 development environment utilities & documentation

anuejn:


doc/development_setup.md
```diff
@@ -0,0 +1,34 @@
+# Development Setup
+
+This document should get you up to speed and give you a working development environment for
+`transcribee`. The instructions in this document may not be used for production installations!
+
+The transcribee project uses a development-setup managed by `nix`. This allows us to all use the
+same software & versions conviniently. Thus, [you need to have a working nix installation on your
```
phlmn, line 7: This is a pretty big requirement... I think it would be feasible to install 7 dependencies by hand.

doc/development_setup.md
```diff
@@ -0,0 +1,34 @@
...
+system](https://nix.dev/tutorials/install-nix).
+
+## Minimal setup
+
+If you just want to try out `transcribee`, you need to go through the following steps:
+
+1. [Install `nix` on your system](https://nix.dev/tutorials/install-nix).
```
phlmn, line 14: This is ~10x worse than "just install docker"? :D I think we should mark this as subject to change.

dev.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env nix-shell
+#!nix-shell shell.nix -i bash
```
phlmn, line 2: pretty nifty script! can we support this outside of a nix shell?

dev.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Installs the dependencies of all transcribee components & starts them for development purposes.
+# This script should work on a fresh clone of transcribee & should only require nix.
+
+set -euxo pipefail
+
+echo -e "\033[1m# setting up backend:\033[0m\n"
+pdm install -p backend/
+pdm run -p backend/ manage migrate
+pdm run -p backend/ manage create_superuser_if_not_exists --user test --pass test
```
pajowu, line 12: I'd vote for `admin:admin` instead of `test:test`

anuejn: maybe a combination, to signify that it is a test user?

pajowu (approved):

phlmn (approved):

PR #44, opened 2023-03-16
Automerge, actual live transcription

pajowu:


pajowu:
This is currently a stub pr, I'll update it with the code I already have later

worker/run.py
```diff
@@ -21,11 +21,18 @@ async def main():
         help="url to the task coordinator (aka the transcribee backend)",
         default="http://localhost:8000",
     )
+    parser.add_argument(
```
anuejn, line 24: why is this seperate?

anuejn: and can I have a default that picks the coordinator?

anuejn (commented):
Seems good to me :)

anuejn (approved):

PR #46, opened 2023-03-17
📦 Integrate Worker into dev setup

anuejn:
- 🫧 backend: make last_seen of Worker nullable
- 🖨️ make worker logging more silent
- 📦 add worker to dev setup


backend/transcribee_backend/base/models.py
```diff
@@ -39,7 +39,7 @@ class Worker(models.Model):
     token = models.CharField(max_length=128, default=generate_worker_token)
     name = models.CharField(max_length=255, null=False, blank=False)
 
-    last_seen = models.DateTimeField()
+    last_seen = models.DateTimeField(default=None, blank=True, null=True)
```
pajowu, line 42: I don’t think we need default=None here

pajowu (approved):
I small comment, LGTM otherwise

PR #49, opened 2023-03-24
Diarization & Aligment Worker

pajowu:


pajowu:
This PR is based on the changes from #44, that's why the PR base is that branch. I will update the pr base branch once #44 is merged into main

backend/transcribee_backend/base/admin.py
```diff
@@ -2,9 +2,27 @@
 
 from .models import Document, DocumentUpdate, Task, User, Worker
 
+
+@admin.action(description="Requeue task")
```
anuejn, line 6: Handy :)

frontend/src/editor/DebugPanel.tsx
```diff
@@ -54,7 +54,7 @@ function Button(props: JSX.IntrinsicElements['button']) {
 const exampleElement: Element = {
   type: 'paragraph',
   lang: 'de',
-  speaker: 'Speaker 1',
+  speakers: [0],
```
anuejn, line 57: what does this do?

frontend/src/editor/types.ts
```diff
@@ -8,12 +8,13 @@ export type Text = {
 export type Paragraph = {
   type: 'paragraph';
   children: Text[];
-  speaker: string;
+  speakers: number[];
```
anuejn, line 11: why / how can a paragraph have multiple speakers?

worker/README.md
```diff
@@ -5,11 +5,30 @@ a transcription worker for the transcribee open-source transcription platform
...
+See .env.example for an example of how this file could look.
+
+The most important settings are explained here:
+
+The worker needs multiple models, which will automatically be downloaded if they to not exist yet.
+They will by default be stored in `./.data/models`.
+You can change this directory by setting `MODELS_DIR`
+
+The diarization models are downloaded from huggingface.
+You need to accept the terms of the models first on [this page](https://hf.co/pyannote/speaker-diarization) and [this page](https://hf.co/pyannote/segmentation).
```
anuejn, line 31: Huh, this really sucks. Do they host the models somewhere else?

worker/pyproject.toml
```diff
@@ -13,24 +13,34 @@ dependencies = [
...
-    "torchaudio>=0.13.1",
-    "torch>=1.13.1",
-    "automerge @ git+https://github.com/transcribee/automerge-py.git@rewrite",
+    "torchaudio==0.11.0",
+    "torch==1.11.0",
+    "automerge @ git+https://github.com/transcribee/automerge-py.git@c87d3b39bb271a3b9efcee7aba059b432cf6c09b",
     "websockets>=10.4",
+    "pyannote-audio>=2.1",
+    "torchvision==0.12.0",
+    "hmmlearn @ git+https://github.com/hmmlearn/hmmlearn@e22127765b0ca70d01976df2a369667756b8445e",             # hmmlearn version which actually compiled 🙃. Needs to be specified here instead of in pdm.resolution.overrides because pdm crashes otherwise 🙃🙃
```
anuejn, line 25: oof

frontend/src/editor/types.ts
pajowu: Short (unsurprising) answer: There might be more than one person speaking in a paragraph.

Short definition of terminology: A Segment is the output of diarization. It has a start, end and a list of speaker ids.

A Paragraph is the output of transcription. It has a start, end, list of speakers and text children.

The reason to change this here is: The segments and paragraphs do not line up perfectly. There might be multiple segments overlapping one paragraph. The question is how we deal with that. One option would be to split the paragraph to align with the segmetns – but then the same problem occurs with single words (a segment might end/start during a word).

The way the code currently deals with that is: A paragraph gets assigned the speakers of all segments that overlap it. I think this needs more finetuning in the future (for example splitting paragraphs in some cases), but is a good baseline for now.

worker/README.md
pajowu: Nope, this is a choice the pyannote developers made. huggingface offers public download by default. We could ask the developers to open the downloads again. However I think that doing it this way for now is okay. In the future I think we could offer a download script that takes all needed tokens and a bit of config and downloads all needed models into one directory. This would allow an easier deployment

worker/pyproject.toml
pajowu: !!. I spend way to much time debugging pdm for these PRs

frontend/src/editor/types.ts
pajowu: As discussed: I changed the types to have an optional `speaker` (assigned if there is only one possible speaker) and a list of `alternative_speakers`

worker/transcribee_worker/diarize.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def diarize(audio, progress_callback) -> List[Segment]:
+    pipeline = Pipeline.from_pretrained(
+        "pyannote/speaker-diarization@2.1",
+        use_auth_token=settings.HUGGINGFACE_TOKEN,
+        cache_dir=settings.MODELS_DIR,
+    )
+    device_name = "cpu"
+    if torch.backends.mps.is_available():
```
anuejn, line 33: we probably want to do this for other backends as well in the future?

worker/transcribee_worker/whisper_transcribe.py
```diff
@@ -109,13 +118,14 @@ def handle_new_segment(
...
                         end=end * 10,
+                        conf_ts=conf_ts,
                     )
                 )
 
             paragraph = Paragraph(
                 children=atoms,
-                speaker=UNKNOWN_SPEAKER,
-                lang=lang_code,
+                alternative_speakers=[],
```
anuejn, line 127: why no default for this?

worker/transcribee_worker/whisper_transcribe.py
```diff
@@ -193,3 +202,50 @@ async def transcribe(
 
     for task in pending:
         task.cancel()
+
+
+async def recombine_split_words(
```
anuejn, line 207: maybe we want to also recombine sentences in the future? so that paragraph breaks only occur on sentence end symbols like `.,?!`

worker/transcribee_worker/worker.py
```diff
@@ -59,13 +70,26 @@ def _get_tmpfile(self, filename: str) -> Path:
...
         response = requests.get(file_url)
-        return BytesIO(response.content)
+        return response.content
+
+    def get_document_audio(self, document: ApiDocument) -> Optional[BytesIO]:
+        b = self.get_document_audio_bytes(document=document)
+        if b is not None:
+            return BytesIO(b)
+
+    def get_document_audio_path(self, document: ApiDocument) -> Optional[Path]:
```
anuejn, line 86: for what do we need that?

anuejn (approved):
Looks good. We can discuss / resolve the things I commented in the future :)

worker/transcribee_worker/diarize.py
pajowu: Yes, but I can't really test this, so I decided to just use the default. I think we probably want to config option for this in the long run

PR #52, opened 2023-03-30
✨ various UI improvements

anuejn:
Try it :)

PR #53, opened 2023-03-30
♫ add playback bar

anuejn:
♫🔊🎤🎶

### DIFF
--- a/backend/transcribee_backend/base/views.py
+++ b/backend/transcribee_backend/base/views.py
@@ -9,4 +9,6 @@ def serve_signed(request, *args, salt: str, max_age: int, **kwargs) -> FileRespo
     path = request.get_full_path()
     if not verify_url(path, salt=salt, max_age=max_age):
         raise PermissionDenied()
-    return serve(request, *args, **kwargs)
+    response = serve(request, *args, **kwargs)
+    response["Access-Control-Allow-Origin"] = "*"
+    return response
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -30,6 +30,7 @@
     "@tailwindcss/forms": "^0.5.3",
     "@types/react": "^18.0.28",
     "@types/react-dom": "^18.0.11",
+    "@types/wavesurfer.js": "^6.0.3",
     "@typescript-eslint/eslint-plugin": "^5.54.0",
     "@typescript-eslint/parser": "^5.54.0",
     "autoprefixer": "^10.4.13",
@@ -65,6 +66,8 @@
     "tailwindcss": "^3.2.7",
     "vite-plugin-top-level-await": "^1.3.0",
     "vite-plugin-wasm": "^3.2.2",
+    "wavesurfer-react": "^2.2.2",
+    "wavesurfer.js": "^6.6.2",
     "wouter": "^2.10.0",
     "yjs": "^13.5.48"
   },
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -11,6 +11,7 @@ specifiers:
   '@tailwindcss/forms': ^0.5.3
   '@types/react': ^18.0.28
   '@types/react-dom': ^18.0.11
+  '@types/wavesurfer.js': ^6.0.3
   '@typescript-eslint/eslint-plugin': ^5.54.0
   '@typescript-eslint/parser': ^5.54.0
   autoprefixer: ^10.4.13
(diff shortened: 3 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -73,7 +73,7 @@ repos:
     hooks:
       - id: format
         name: format
-        entry: pnpm --prefix frontend/ run check:format
+        entry: pnpm --prefix frontend/ run format
```
pajowu, line 76: 😍

frontend/src/App.tsx
```diff
@@ -1,24 +0,0 @@
-import { Route, Router, Switch } from 'wouter';
```
pajowu, line 1: Why do we lowercase all filesnames?

frontend/src/components/popup.tsx
```diff
@@ -0,0 +1,54 @@
...
+      {show ? (
+        <div
+          className={clsx(
+            'p-4',
+            'bg-white',
+            'border-black',
+            'border-2',
+            'shadow-brutal',
+            'rounded-lg',
+            'm-2',
```
pajowu, line 41: Popper does not like margins on itself (I think this line is was causes it). The popper element reports in the console:

> CSS "margin" styles cannot be used to apply padding between the popper and its reference element or boundary. To replicate margin, use the `offset` modifier, as well as the `padding` option in the `preventOverflow` and `flip` modifiers. 

More Details: https://popper.js.org/docs/v2/migration-guide/#4-remove-all-css-margins

frontend/src/pages/new_document.tsx
```diff
@@ -48,11 +47,7 @@ export default function NewDocumentPage() {
   };
 
   return (
-    <div
-      className="h-screen p-6 flex items-center justify-center"
-      onDragOver={(e) => e.preventDefault()}
-      onDrop={(e) => e.preventDefault()}
-    >
+    <AppCenter onDragOver={(e) => e.preventDefault()} onDrop={(e) => e.preventDefault()}>
       <Dialog>
```
pajowu, line 51: Can I have a back button for the new document dialog too?

frontend/src/pages/user_home.tsx
```diff
@@ -52,7 +56,30 @@ export default function HomePage() {
...
+              'border',
+              'border-gray-200',
+              'hover:shadow-lg',
+              'hover:scale-105',
+              'transition-all',
+              'flex',
+            )}
+          >
+            {' '}
+            <AiOutlinePlus size="1.75rem" />
```
pajowu, line 79: I think the aria label (or `visually-hidden` span) is missing here for accessibility.

frontend/src/common/top_bar.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { IconButton, PrimaryButton } from '../components/button';
+import { Popup } from '../components/popup';
+
+export const TopBar = primitiveWithClassname('div', 'mb-8 flex items-center gap-4 justify-between');
+export const TopBarTitle = primitiveWithClassname('h2', 'text-xl font-bold');
+export const TopBarPart = primitiveWithClassname('div', 'gap-4 flex items-center');
+
+export function MeButton() {
+  return (
+    <Popup button={<IconButton icon={BiUser} />}>
```
pajowu, line 15: This is missing an aria label / visually hidden span

pajowu (changes requested):
Looks good in general, just some smaller changes (I think most of them are from #52)

frontend/src/pages/new_document.tsx
phlmn: Maybe a cancel button instead?

frontend/src/components/popup.tsx
anuejn: ah good catch!

frontend/src/App.tsx
pajowu: Nope, was just wondering because I always found it intuitive to have the main export as the filename

pajowu:
I noticed that the player bar waveform is not shown until the editor state is fully loaded. The cause seems to be that both editor state processing and waveform generation are done on the main thread. It might be a good idea to move this to a webworker or similar (see also https://github.com/wavesurfer-js/wavesurfer.js/issues/1867 ). But that doesn't have to happen in this PR

pajowu (approved):

anuejn:
Sounds good :). I do think though, that we want both the wavesurfer loading as well as the automerge merging to happen in a webworker. But lets do that in the future :)

PR #59, opened 2023-04-03
✨ Reencode task for worker

pajowu:
Reencode everything to 128k cbr mp3 and m4a

### DIFF
--- a/backend/transcribee_backend/routers/document.py
+++ b/backend/transcribee_backend/routers/document.py
@@ -17,6 +17,7 @@ from fastapi import (
 )
 from sqlmodel import Session, select
 from transcribee_backend.auth import (
+    get_authorized_worker,
     validate_user_authorization,
     validate_worker_authorization,
 )
@@ -33,6 +34,7 @@ from ..models import (
     Task,
     TaskType,
     UserToken,
+    Worker,
 )
 from .user import get_user_token
 
@@ -44,14 +46,20 @@ def now_tz_aware() -> datetime.datetime:
 
 
 def create_default_tasks_for_document(session: Session, document: Document):
+    reencode_task = Task(
+        task_type=TaskType.REENCODE,
+        task_parameters={},
+        document_id=document.id,
+    )
+    session.add(reencode_task)
+
     transcribe_task = Task(
         task_type=TaskType.TRANSCRIBE,
         task_parameters={"lang": "auto", "model": "base"},
         document_id=document.id,
+        dependencies=[reencode_task],
     )
     session.add(transcribe_task)
-    session.commit()
-    session.refresh(transcribe_task)
 
     align_task = Task(
         task_type=TaskType.ALIGN,
@@ -60,8 +68,6 @@ def create_default_tasks_for_document(session: Session, document: Document):
         dependencies=[transcribe_task],
     )
     session.add(align_task)
-    session.commit()
-    session.refresh(align_task)
 
     speaker_identification_task = Task(
         task_type=TaskType.IDENTIFY_SPEAKERS,
@@ -70,8 +76,6 @@ def create_default_tasks_for_document(session: Session, document: Document):
         dependencies=[align_task],
     )
     session.add(speaker_identification_task)
-    session.commit()
-    session.refresh(speaker_identification_task)
 
 
 @document_router.post("/")
@@ -89,8 +93,6 @@ async def create_document(
     )
 
     session.add(document)
-    session.commit()
-    session.refresh(document)
 
     stored_file = media_storage.store_file(file.file)
     file.file.seek(0)
@@ -104,15 +106,13 @@ async def create_document(
     )
 
     session.add(media_file)
-    session.commit()
-    session.refresh(media_file)
 
     tag = DocumentMediaTag(media_file_id=media_file.id, tag="original")
     session.add(tag)
-    session.commit()
-    session.refresh(tag)
 
     create_default_tasks_for_document(session, document)
+
+    session.commit()
     return document.as_api_document()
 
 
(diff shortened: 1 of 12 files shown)


anuejn:
Do we really want to do this in the backend?.
I have the vague recollection, that we discussed to do it in the frontend, because it saves us compute and the users bandwidth.

pajowu:
Yes and no. I think it makes sense to do this in the frontend too, but I think we should have to option in the backend as well. a) to make sure the file is in the right format and b) to handle cases where the conversion in the frontend is not possible (e.g. files larger than 2GB, browsers without wasm)

worker/pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "torchvision==0.14.1",
     "hmmlearn @ git+https://github.com/hmmlearn/hmmlearn@e22127765b0ca70d01976df2a369667756b8445e",             # hmmlearn version which actually compiled 🙃. Needs to be specified here instead of in pdm.resolution.overrides because pdm crashes otherwise 🙃🙃
     "lightning-lite>=1.8.6",
+    "ffmpeg-python>=0.2.0",
```
anuejn, line 27: could you add ffmpeg to our shell.nix or does this bring its own ffmpeg?

anuejn (commented):
I am fine with merging transcription at the worker since it is an improvement from the status quo :). However I would like to investigate transcoding at the frontend in the future.

anuejn:
fair points :)

anuejn:
Ah we probably also need a way to inform the frontend, that the audio file changed. -> Future

anuejn (approved):

backend/transcribee_backend/routers/document.py
```diff
@@ -57,6 +69,7 @@ def create_default_tasks_for_document(session: Session, document: Document):
         task_type=TaskType.TRANSCRIBE,
         task_parameters={"lang": "auto", "model": "base"},
         document_id=document.id,
+        dependencies=[reencode_task],
```
anuejn, line 72: is this needed?

backend/transcribee_backend/routers/document.py
```diff
@@ -191,3 +204,47 @@ async def websocket_endpoint(
...
+    tags: list[str] = Form(),
+    file: UploadFile = File(),
+    session: Session = Depends(get_session),
+    authorized_worker: Worker = Depends(get_authorized_worker),
+) -> ApiDocument:
+    statement = select(Task).where(
+        Task.document_id == document_id, Task.task_type == TaskType.REENCODE
+    )
+    task = session.exec(statement).one_or_none()
+    if task is None:
```
anuejn, line 221: for what do we need the task here?

frontend/src/editor/player.tsx
```diff
@@ -18,7 +18,16 @@ export function PlayerBar({
   documentContent: Descendant[];
 }) {
   const { data } = useGetDocument({ document_id: documentId });
-  const audioFile = data?.media_files[0]?.url;
+  let audioFile = data?.media_files[0]?.url;
+  const audioElement = document.createElement('audio');
+  data?.media_files.forEach((media_file) => {
+    if (
+      media_file.tags.indexOf('original') == -1 &&
+      audioElement.canPlayType(media_file.content_type) == 'probably'
```
anuejn, line 26: nice :D

backend/transcribee_backend/routers/document.py
pajowu: Yes, `load_audio` only supports a small number of formats (eg doesn’t do mp4) -> ffmpeg converts them to a safe format

backend/transcribee_backend/routers/document.py
pajowu: This is a permission check: the worker is only allowed to upload media if it is assigned to the reencode task

backend/transcribee_backend/routers/document.py
anuejn: Hm... can we skip the dependency if we are likely to be able to load it directly? Or use a library that can load more formats? This seems like a downgrade performance-wise :/

pajowu: I'd prefer not to, as it also calms my mind a bit to know that (after it went through ffmpeg, which is it list fuzzed a bit), the files are in a sortof-safe state. Also the reencoding is incredibly fast

backend/transcribee_backend/routers/document.py
```diff
@@ -191,3 +204,47 @@ async def websocket_endpoint(
...
+        created_at=now_tz_aware(),
+        changed_at=now_tz_aware(),
+        document_id=task.document_id,
+        file=stored_file,
+        content_type=magic.from_descriptor(file.file.fileno(), mime=True),
+    )
+
+    session.add(media_file)
+    session.commit()
+    session.refresh(media_file)
```
rroohhh, line 240: This refresh is also unnecessary. Doesn't `sqlmodel` automatically expire on commit and then fetch the most recent data when its used later?

rroohhh (changes requested):

backend/transcribee_backend/routers/document.py
pajowu: Nice, thanks for all the three comments. I think I just commited the `add`/`commit`/`refresh`-code somewhere from the sqlmodel docs without thinking about it again

rroohhh (approved):
`load_audio` should probably be moved to the `ffmpeg-python` package at some point aswell

rroohhh (approved):
`load_audio` should probably be moved to the `ffmpeg-python` package at some point aswell

PR #63, opened 2023-04-05
Cache docker build & fix host key verification error

pajowu:
The docker build still doesn't work after this because whispercpps bazel fuckery doesn't run on nixos


anuejn (approved):
looks good :)

PR #66, opened 2023-04-13
👷 Use dev script for installing dependencies in ci

pajowu:


packaging/dev_prepare.sh
```diff
@@ -1,19 +1,19 @@
 #!/usr/bin/env bash
```
phlmn, line 1: Should we rename this to something like `install_dependencies.sh`?

phlmn:
Wow, installing the worker dependencies is slow...

anuejn (approved):
looks good :)

PR #67, opened 2023-04-17
✨ Add webvtt export

pajowu:


anuejn (approved):
Looks good :).
I would move the button somewhere else in the future but for now at the bottom seems like a fair place to me :)

PR #69, opened 2023-04-17
nix: Use stdenv without CC on macOS

phlmn:
Actually not sure if this is a good idea...

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -18,8 +18,12 @@ let
     doCheck = false;
   });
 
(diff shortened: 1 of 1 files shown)


anuejn:
still does not work for me :/

pajowu:
Doesn't this just cause the build system to use the compiler provided by mac os (if no compiler is configured in the nix system packages?)

anuejn:
Can we close this @phlmn

pajowu:
Closing as it seems to be fixed with #77

PR #71, opened 2023-04-25
✨ Add basic collaborative editing with automerge

phlmn:


anuejn (approved):
Works like a charm :)

phlmn:
> Works like a charm :)

Nice :) 

But I don't get why the CI thinks the lockfile is outdated...

frontend/pnpm-lock.yaml
```diff
@@ -1,289 +1,329 @@
...
-  wouter: 2.10.0_react@18.2.0
+  '@audapolis/webvtt-writer':
+    specifier: ^1.0.6
+    version: 1.0.6
+  '@automerge/automerge':
+    specifier: file:/Users/anuejn/Developer/transcribee/frontend/deps/automerge-2.0.2.tgz
+    version: file:deps/automerge-2.0.2.tgz
+  '@automerge/automerge-wasm':
+    specifier: file:/Users/anuejn/Developer/transcribee/frontend/deps/automerge-wasm-0.1.25.tgz
+    version: file:deps/automerge-wasm-0.1.25.tgz
```
phlmn, line 16: Absolute paths don't seem like a good idea. Removing the overrides results in relative paths... Is this a pnpm bug?

anuejn: almost certainly

anuejn: it works with absolute paths but not with relative ones

anuejn: I think we need to use `pnpm install --no-frozen-lockfile` as ugly as it is

phlmn: Why do we need the overrides anyway?

anuejn: because they depend on each other?

pajowu: I think we can now just switch to automerge@2.0.3, which should contain all fixes we need. This also means we can get rid of the tar.gz in the repo

anuejn: 🥳

frontend/src/editor/automerge_websocket_provider.ts
```diff
@@ -1,4 +1,5 @@
 import { Observable } from 'lib0/observable';
```
pajowu, line 1: `lib0` is still an old dependency from the yjs-devs. Does it make since to keep it now that we decided against yjs or can we switch this to something from the stdlib or self-built?

frontend/src/editor/slate_types.d.ts
```diff
@@ -1,10 +1,11 @@
 import { BaseEditor } from 'slate';
 import { ReactEditor } from 'slate-react';
 import { Paragraph, Text } from './types';
+import { AutomergeEditor } from 'slate-automerge-doc';
 
 declare module 'slate' {
   interface CustomTypes {
-    Editor: BaseEditor & ReactEditor;
+    Editor: BaseEditor & ReactEditor & AutomergeEditor;
```
pajowu, line 8: `AutomergeEditor` is already `BaseEditor & ...`, so the following should work?

```suggestion
    Editor: AutomergeEditor & ReactEditor;
```

pajowu (approved):
Works great. LGTM, just two inline thoughts.

phlmn:
Seems like the lockfile of slate-automerge-doc was simply not up to date 🙈 Let's cross fingers the CI works now 🤞

PR #74, opened 2023-04-25
👷 Cache pypackages in ci

pajowu:


anuejn:
what kind of rust stuff are we building in the ci?

pajowu:
automerge. My hope is that this would make the macos build less flaky. I don't fully understand why it breaks sometimes yet. Seems like somehow cargo sometimes fails to find the correct commit???

anuejn:
would't be the better approach to build wheels for our automerge bindings?

pajowu:
Maybe. My hope was that this give a good-enough hotfix until the automerge developers release a new version with the fixes so we can just switch back to upsteam

anuejn:
fair :D

anuejn (approved):

PR #79, opened 2023-04-26
⬆️ Upgrade cache action to v3

pajowu:
V2 uses node12, which GitHub deprecated and will be remove in september: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/


anuejn (approved):
LGTM

PR #86, opened 2023-04-27
use torch cpu version

rroohhh:
otherwise the worker will not start on linux systems without a cuda installation


anuejn (approved):
looks good :)

PR #110, opened 2023-04-28
👷‍♀️ switch to macos-xl runners

anuejn:


anuejn:
Okay, it seems like there are not enough macos-latest-xl runners at the moment to pick up the tasks

PR #117, opened 2023-04-29
⚗️ improve nix install speed?

pajowu:


pajowu:
The new install action uses single-user install. This seems to be slightly faster than the cachix install action used before

anuejn (approved):
Nice :)

PR #122, opened 2023-05-02
Misc fixes

rroohhh:


anuejn (approved):
looks good :)

rroohhh:
This now switches to the fastapi security stuff, which is also integrated with openapi, but `openapi-typescript-fetch` doesn't support that I think.

So maybe we want to build support for this sometime.

anuejn:
hm... lets not use it then? sounds a bit cumbersome for not a lot of gain

rroohhh:
Maybe, but apart from the `openapi-typescript-fetch` thing this is a lot more proper.

backend/transcribee_backend/auth.py
```diff
@@ -66,11 +65,19 @@ def validate_user_authorization(session: Session, authorization: str):
...
-):
-    return validate_user_authorization(session, authorization)
+class TokenAuth(HTTPBase):
+    def __init__(self):
+        self.model = HTTPBaseModel(scheme="token")
+        self.scheme_name = "Token"
+
+    async def __call__(
+        self, request: Request, session: Session = Depends(get_session)
+    ) -> Optional[UserToken]:
```
pajowu, line 75: Is this really `Optional`? I think we always return a `UserToken`, otherwise we raise an Exception

pajowu:
I'm not sure about fastapi.security as well, as long as our frontend (the only place we really need this) doesn't work with it? And if we switch, should we switch the worker auth as well?

pajowu:
The other stuff looks good to me

rroohhh:
I dropped the fastapi.security stuff and moved it to #140

backend/transcribee_backend/auth.py
```diff
@@ -50,8 +50,8 @@ def validate_user_authorization(session: Session, authorization: str):
         raise HTTPException(status_code=401)
 
     try:
-        token_data = b64decode(token.encode()).decode()
```
pajowu, line 53: Why this?

rroohhh: `b64decode` already internally encodes ascii strings, so this seems unnecessary?

PR #126, opened 2023-05-03
🚸 Set highlight if clicked outside of word

### LINKED ISSUE #120: Playback position not moved when clicking left of text
When clicking left of the transcript text (but right of the speaker name column), the text-cursor is placed at the start of the transcript line. However the playback position is not moved to the same point.

https://user-images.githubusercontent.com/6545097/235660702-fbe00e12-8652-471d-9171-e5edf9804704.mov

pajowu:
Fixes #120

### DIFF
--- a/frontend/src/editor/transcription_editor.tsx
+++ b/frontend/src/editor/transcription_editor.tsx
@@ -154,6 +154,15 @@ export function TranscriptionEditor({ documentId }: { documentId: string }) {
           <Editable
             renderElement={(props) => renderElement(props, editor.doc)}
             renderLeaf={renderLeaf}
+            onClick={() => {
+              const selection = document.getSelection();
+              if (
+                selection?.isCollapsed &&
+                selection.anchorNode?.parentElement?.parentElement?.classList.contains('word')
+              ) {
+                selection.anchorNode.parentElement.click();
+              }
+            }}
           />
         </Slate>
       </div>


frontend/src/editor/transcription_editor.tsx
```diff
@@ -48,7 +50,40 @@ function renderElement(
...
+                diff = clientX - bbox.right;
+              } else {
+                diff = 0;
+              }
+              if (diff < nearest_diff) {
+                nearest_diff = diff;
+                nearest = i;
+              }
+            }
+            window.dispatchEvent(new TextClickEvent(element.children[nearest]));
```
anuejn, line 84: can we also remove the handler in 109 following then?

pajowu: Nope, this doesn't work once you have multiline text. maybe we should just handle that cleanly here too and then get rid of 109

anuejn: I would prefer that (I also think, that this might be faster than having many listeners)

pajowu: I rewrote it to rather use the `selection` api and click the correct word

pajowu: Okay, after a bit more investigation: I think it's impossible to perfectly replicate the caret-behaviour in the onclick handler. I instead opted to use the `getSelection`-API in an onclick handler

anuejn (approved):
even if the code is a bit unholy: it works like a charm :)

PR #136, opened 2023-05-07
🌚🌝 Dark Mode

pajowu:


pajowu:
This is a very hard dark mode, using full black as the main color. Should we use something softer like `gray-900`?

anuejn:
This is a bit to hard for my taste. I would vote for the following changes: 
* [ ] neutral-900 as a background.
* [ ]  darker-than-white borders for dialogs and buttons.
* [ ] lighter blue border for active input fields
* [ ] I do not like the blue-ish tint of the slate colors for the shadows

pajowu:
I implemented them all

anuejn (approved):

PR #140, opened 2023-05-09
Switch authentication to fastapi.security

rroohhh:
We probably  want https://github.com/transcribee/openapi-typescript-fetch support before merging this.


anuejn:
If somebody is motivated to implement support in typscript-openapi-fetch for this, we can open this again

PR #142, opened 2023-05-09
🙈 Re-add new backend db to gitignore

pajowu:


anuejn:
why keep both? also I have some more fixes incoming 1 mom

pajowu:
Why not?

anuejn:
I would also include the storage directory properly like in https://github.com/transcribee/transcribee/pull/119/commits/3d55afa40a3e78544cb6c08ab984c532bb417245

anuejn (approved):

PR #161, opened 2023-05-12
✨ Show timestamps in webvtt format

pajowu:


frontend/src/editor/transcription_editor.tsx
```diff
@@ -16,6 +16,18 @@ const LazyDebugPanel = lazy(() =>
   import('./debug_panel').then((module) => ({ default: module.DebugPanel })),
 );
 
+function formatTimestamp(time: number): string {
+  const seconds = time % 60;
+  time = (time - seconds) / 60;
+  const minutes = time % 60;
+  const hours = (time - minutes) / 60;
+  if (hours > 0) {
```
rroohhh, line 24: I don't know what the convention is, but maybe this check should take the total length into account?

rroohhh (approved):

frontend/src/editor/transcription_editor.tsx
pajowu: Ah wait, you're right. This should check the total length of the audio, not just the current timestamp.

pajowu: Once the `reencode_worker` branch is landed, we can add this. At the moment, this isn't easily possible

pajowu: Once the `pajowu/stats` branch (#144) is merged, we can easily to that

PR #162, opened 2023-05-14
✨ Add speaker names and word timings to webvtt export

pajowu:
Code based on the identical feature in audapolis

### DIFF
--- a/frontend/src/components/form.tsx
+++ b/frontend/src/components/form.tsx
@@ -1,16 +1,25 @@
 import { ReactNode } from 'react';
 import { primitiveWithClassname } from '../styled';
+import clsx from 'clsx';
 
 export type FormControlProps = {
   label: string;
   error?: string;
   children: ReactNode;
+  disabled?: boolean;
 };
 
-export function FormControl({ label, error, children }: FormControlProps) {
+export function FormControl({ label, error, disabled, children }: FormControlProps) {
   return (
     <label className="block">
-      <span className="text-sm font-medium">{label}</span>
+      <span
+        className={clsx(
+          'text-sm font-medium',
+          disabled ? 'text-slate-400 dark:text-neutral-500' : '',
+        )}
+      >
+        {label}
+      </span>
       {children}
       {error && <p className="text-red-600 text-sm mt-1">{error}</p>}
     </label>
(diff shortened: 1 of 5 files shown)


frontend/src/components/form.tsx
```diff
@@ -1,16 +1,25 @@
...
 
 export type FormControlProps = {
   label: string;
   error?: string;
   children: ReactNode;
+  disabled?: boolean;
 };
 
-export function FormControl({ label, error, children }: FormControlProps) {
+export function FormControl({ label, error, disabled, children }: FormControlProps) {
```
anuejn, line 12: should we not do something to actually disable the control here?

frontend/src/components/form.tsx
```diff
@@ -27,8 +36,43 @@ export const Input = primitiveWithClassname('input', [
   'mt-0.5',
   'dark:bg-neutral-900',
   'dark:focus:ring-blue-400 dark:focus:border-blue-400',
+  'disabled:border-slate-400 disabled:dark:border-neutral-500 disabled:text-slate-400 disabled:dark:text-neutral-500',
 ]);
 
+export function Checkbox({
```
anuejn, line 42: why do we not use FormControl to implement this?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -161,10 +162,10 @@ export function TranscriptionEditor({ documentId }: { documentId: string }) {
...
         <SecondaryButton
           className="my-4"
-          onClick={() => {
-            const vtt = generateWebVtt(Automerge.toJS(editor.doc));
-            downloadTextAsFile('document.vtt', 'text/vtt', vtt.toString());
-          }}
+          onClick={() =>
+            showModal(<WebvttExportModal editor={editor} onClose={() => showModal(null)} />)
+          }
+          disabled={editor.doc.children === undefined || !canGenerateVtt(editor.doc.children)}
```
anuejn, line 168: I think this is kinda unintuitive. I would whish for at least a hover tooltip why this is disabled

anuejn: or make it selectable and then display a text that it is currently impossible due to missing timings

anuejn (changes requested):

anuejn (approved):

PR #164, opened 2023-05-16
✨ Allow deleting a document

pajowu:


pajowu:
I'll replace the `confirm` dialog once #160 is merged, which adds a modal component

backend/transcribee_backend/routers/document.py
```diff
@@ -147,6 +151,45 @@ def get_document(
...
+        execution_options={"synchronize_session": "fetch"},
+    )
+
+    session.execute(delete(Task).where(Task.document == document))
+
+    session.delete(document)
+    session.commit()
+
+    for path in paths_to_delete:
+        os.remove(path)
```
anuejn, line 189: If something goes wrong here (i.e. the file does not exist) we get an exception but the document is still gone.
further files will not be deleted and we dont have a handle to them anymore. this is kinda bad?
maybe we want to delete the document as a very last step and collect errors here instead of directly throwing?

anuejn: Indicating to the user that the document is gone, when in fact it is not fully feels bad

rroohhh, line 189:
```suggestion
        path.unlink()
```

rroohhh (approved):

backend/transcribee_backend/routers/document.py
```diff
@@ -147,6 +151,45 @@ def get_document(
...
+        delete(TaskDependency).where(
+            or_(
+                col(TaskDependency.dependant_on_id).in_(task_ids),
+                col(TaskDependency.dependent_task_id).in_(task_ids),
+            )
+        ),
+        execution_options={"synchronize_session": "fetch"},
+    )
+
+    session.execute(delete(Task).where(Task.document == document))
```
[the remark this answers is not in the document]
pajowu: In theory: Yes. But setting this trough sqlmodel is harder than you'd think. I can try again

backend/transcribee_backend/routers/document.py
pajowu: Catching errors here is a good idea, but I don't think keeping the document around if the deletion fails is a good idea. This would lead to a VERY inconsistent state. I'd prefer something like a cronjob, that deletes files still lying around

backend/transcribee_backend/routers/document.py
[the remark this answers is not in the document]
pajowu: AAARRRGGGHHH. Took me a few hours to realize that sqlite does not honor `ON DELETE` by default. Now it works just as expected 🙃

PR #169, opened 2023-05-17
🐬 add color indicators for speakers

anuejn:


frontend/src/editor/speaker_colors.tsx
```diff
@@ -0,0 +1,29 @@
...
+  editor,
+}: {
+  children: JSX.Element;
+  editor: Editor;
+}) {
+  const [colors, setColors] = useState<Record<string, string>>({});
+  useEffect(() => {
+    if (editor?.doc?.speaker_names) {
+      const colors = Object.fromEntries(
+        Object.entries(editor?.doc?.speaker_names).map(([k, _v], i) => [k, getColor(i)]),
```
pajowu, line 22: Why do we only show these for speakers which have a name assigned?

pajowu, line 22: Is it a good idea that this is based on the name and not on the id?

frontend/src/editor/speaker_colors.tsx
```diff
@@ -0,0 +1,29 @@
...
+  children: JSX.Element;
+  editor: Editor;
+}) {
+  const [colors, setColors] = useState<Record<string, string>>({});
+
+  const paragraphs = editor.doc.children || [];
+  const speakers = paragraphs.map((p) => p.speaker);
+
+  useEffect(() => {
+    const colors = Object.fromEntries([...new Set(speakers)].map((uuid, i) => [uuid, getColor(i)]));
```
rroohhh, line 24: Why not use a mapping from `uuid -> color` (here and later)?
This would avoid corner cases when there are duplicate speaker names.

frontend/src/editor/speaker_colors.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect, useState } from 'react';
+import { Editor } from 'slate';
+
+export const SpeakerColorsContext = React.createContext<Record<string, string>>({});
+
+function getColor(n: number) {
+  const hue = 230 + n * 137.5; // https://en.wikipedia.org/wiki/Golden_angle
+  return `hsl(${hue}, ${Math.min(70 + n * 10, 100)}%, 75%)`;
```
rroohhh, line 8: What do gain by modulating the saturation? 
Maybe we want to just use a hardcoded color progression?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -49,6 +52,10 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
             >
               {formattedTime(startAtom.start)}
             </div>
+            <div
+              className="absolute right-0 top-0 w-2 h-full rounded-md mr-2 -mt-0.5"
+              style={element.speaker ? { backgroundColor: speakerColors[element.speaker] } : {}}
```
rroohhh, line 57: Do we actually want to use `backgroundColor`? How does this interact with the text color? 
What about some kind of colored bar next to the paragraph?

frontend/src/editor/speaker_colors.tsx
anuejn: > What do gain by modulating the saturation?
more diversity in the colors. I thought this way we can generate infinite colors that have kinda optimal distance. Maybe we want a more involved algorithm though. Something like starting with color hashes and then iteratively spreading them out according to perceived color distance?

pajowu:
The padding on this one is off on smaller screens
Screenshot 2023-05-23 at 17 04 08

frontend/src/editor/speaker_colors.tsx
rroohhh: Hmm I am not convinced. Modulating the saturation is only doing something for the first 4 colors. 
I think we would want to use equal luminance colors?
I am still in favor of just a hardcoded list. I don't think having more than say 20 distinct speaker colors will actually help anything.

pajowu (approved):

PR #170, opened 2023-05-17
anuejn/confidence fix

anuejn:
- 💅 use gradient for confidence colors
- 🚧 wip: set confidence to 1 on user correction

### DIFF
--- a/frontend/src/editor/transcription_editor.tsx
+++ b/frontend/src/editor/transcription_editor.tsx
@@ -1,8 +1,8 @@
-import { Editor } from 'slate';
+import { Editor, Transforms } from 'slate';
 import { Slate, Editable, RenderElementProps, RenderLeafProps } from 'slate-react';
 import { SpeakerDropdown } from './speaker_dropdown';
 import { useEvent } from '../utils/use_event';
-import { TextClickEvent } from './types';
+import { Paragraph, TextClickEvent } from './types';
 import { startTimeToClassName } from './player';
 import clsx from 'clsx';
 import { useContext } from 'react';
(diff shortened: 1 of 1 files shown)


pajowu (changes requested):
I'm not a big fan of this. While it it's nice to show the users the confidence in a more fine-grained way, i'm not sure doing it via the text color is the best way. I find it hard to judge the difference between the colors and to figure out the confidence from these.

The current formula for generating the colors also currently generates colors with contrast less than 4.5:1 ([WCAG 2.1 SC 1.4.3](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html))

anuejn:
the audapolis-way of doing things sadly looks a bit ugly because we already use bg-color to highlight the current word. What would you propose?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -63,14 +64,29 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
...
+    classes.push(startTimeToClassName(wordStart.start));
   }
 
+  const systemPrefersDark = useMediaQuery('(prefers-color-scheme: dark)');
+
+  const color = !leaf.conf
+    ? undefined
+    : systemPrefersDark
+    ? `hsl(0, 100%, ${Math.min(leaf.conf * 50 + 50, 100)}%)`
+    : `hsl(0, 100%, ${Math.min((1 - leaf.conf) * 100, 50)}%)`;
```
anuejn, line 88:
```suggestion
    : `hsl(0, 100%, ${Math.min((1 - leaf.conf) * 100, 45)}%)`;
```

if we do this change we get the required contrast ratios for both light and dark

anuejn:
@pajowu is it okay for you now that the contrast ratio ist always more than 4.5:1 ?

pajowu:
I think using fixed buckets could be a better way (<0.7 red, <0.95 darker red, >0.95 black) or something like that? But I'm fine with merging this for now as long as the contrast is fine

pajowu (approved):

PR #174, opened 2023-05-18
Changes for beta deployment

pajowu:
- 🩹 Remove editable transcribee_proto from dev deps
- ➖ Replace react-responsive with custom useMediaQuery impl
- ⚡️ Cache SWR in local storage

### DIFF
--- a/frontend/src/utils/use_media_query.ts
+++ b/frontend/src/utils/use_media_query.ts
@@ -1,0 +1,10 @@
+import { useEffect, useState } from 'react';
+
+export function useMediaQuery(mediaQuery: string) {
+  const [state, setState] = useState(window.matchMedia(mediaQuery).matches);
+  useEffect(() => {
+    const query = window.matchMedia(mediaQuery);
+    query.addEventListener('change', (q) => setState(q.matches));
+  }, []);
+  return state;
+}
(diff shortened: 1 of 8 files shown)


frontend/src/app.tsx
```diff
@@ -8,6 +8,21 @@ import { NewDocumentPage } from './pages/new_document';
 import { DocumentPage } from './pages/document';
 import { PageNotFoundPage } from './pages/page_not_found';
 import { ModalHolder } from './components/modal';
+import { SWRConfig, Cache, State } from 'swr';
+
+function localStorageProvider(): Cache {
```
phlmn, line 13: Do we really want that? This potentialy leaks data to unauthorized sessions. At least we should clear the cache when logging out or beeing logged out.

pajowu: Probably not. I'll remove it. This was meant as a quick fix because the server currently is very slow (every request takes >1s), but we should fix this in the server, not workaround in the client

frontend/src/utils/use_media_query.ts
```diff
@@ -0,0 +1,10 @@
+import { useEffect, useState } from 'react';
+
+export function useMediaQuery(mediaQuery: string) {
```
anuejn, line 3: why do we do this instead of using the lib? -> provide rationale in the commit message plz

pajowu: Done

anuejn (approved):

PR #178, opened 2023-05-21
✨ add playback speed controls

anuejn:


frontend/src/editor/player.tsx
```diff
@@ -160,3 +163,35 @@ export function PlayerBar({ documentId, editor }: { documentId: string; editor:
...
   return `start-${startTime.toFixed(3).replace('.', '')}`;
 }
+
+function PlaybackSpeedDropdown({ onChange }: { onChange: (v: number) => void }) {
+  const possibleRates = [0.5, 0.7, 1.0, 1.2, 1.5, 1.7, 2.0];
+
+  const [value, setValue] = useState(1.0);
+
+  return (
+    <Dropdown
```
pajowu, line 173: IMHO a classic `<select`-attribute would be better here. For example it also provides an indication of the currently selected element, support for pressing <kbd>Esc</kbd>, ...

pajowu (changes requested):

frontend/src/editor/player.tsx
anuejn: good point

frontend/src/editor/player.tsx
```diff
@@ -173,6 +173,8 @@ export function PlayerBar({ documentId, editor }: { documentId: string; editor:
             <WaveForm id="waveform" cursorColor="red" barWidth={2} normalize responsive={true} />
           </WaveSurfer>
         </div>
+
+        <PlaybackSpeedDropdown onChange={(v) => waveSurferRef.current?.setPlaybackRate(v)} />
```
pajowu, line 177: Maybe we should also switch the wavesurfer `backend` to `MediaElement` to the media gets pitch corrected https://github.com/wavesurfer-js/wavesurfer.js/issues/149#issuecomment-310880071

anuejn: sounds like a good idea 👍

frontend/src/utils/use_blur.ts
```diff
@@ -0,0 +1,22 @@
+import { useRef } from 'react';
+import { useEvent } from './use_event';
+
+export function useOnBlur(ref: HTMLElement | null, callback: () => void) {
```
phlmn, line 4: I think there is a better way... Will investigate this later today

frontend/src/editor/player.tsx
```diff
@@ -182,3 +184,33 @@ export function PlayerBar({ documentId, editor }: { documentId: string; editor:
...
+  const [value, setValue] = useState(1.0);
+
+  return (
+    <select
+      value={value.toFixed(1)}
+      className={clsx(
+        'tabular-nums text-sm font-semibold',
+        'p-2 ml-2 rounded-lg',
+        'appearance-none bg-none border-none focus:ring-0',
+        'hover:bg-gray-200',
```
phlmn, line 200: Needs a `dark:text-black` to be readable on my machine

frontend/src/editor/player.tsx
anuejn: Yes!

frontend/src/utils/use_blur.ts
phlmn: discussion moved here: https://github.com/transcribee/transcribee/pull/179#discussion_r1202632624

pajowu (approved):

phlmn (approved):

frontend/src/editor/player.tsx
phlmn: > Does this also fix the jumping cursor position, when the speed is changed?

Seems not to be the case :(

phlmn: @anuejn Just noticed the backend was not changed to `MediaElement` is there a reason for this?

PR #179, opened 2023-05-21
✨ Popup improvements

anuejn:
- ❌ close popups when they use focus
- 💅 animate popups

### DIFF
--- a/frontend/src/utils/use_blur.ts
+++ b/frontend/src/utils/use_blur.ts
@@ -1,20 +1,9 @@
-import { useRef } from 'react';
 import { useEvent } from './use_event';
 
 export function useOnBlur(ref: HTMLElement | null, callback: () => void) {
-  // we need this terrible timeout hack because sometimes click events are fired, that are not on
-  // the target. these are immidiately followed by the actual event
-  const timeout = useRef<number | null>(null);
-
   useEvent<MouseEvent>('click', (e) => {
-    if (ref && e.target) {
-      if (ref.contains(e.target as Node)) {
-        timeout.current && clearTimeout(timeout.current);
-      } else {
-        timeout.current = setTimeout(() => {
-          callback();
-        }, 10);
-      }
+    if (ref && e.target && !ref.contains(e.target as Node)) {
+      callback();
     }
   });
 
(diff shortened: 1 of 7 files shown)


frontend/src/utils/use_blur.ts
```diff
@@ -0,0 +1,22 @@
+import { useRef } from 'react';
+import { useEvent } from './use_event';
+
+export function useOnBlur(ref: HTMLElement | null, callback: () => void) {
+  // we need this terrible timeout hack because sometimes click events are fired, that are not on
+  // the target. these are immidiately followed by the actual event
```
pajowu, line 6: How/when/why does this happen?

anuejn: I am not really sure, why this happens, but it happens whenever I focus the slate editor and then try to operate a dropdown menu for speaker options

frontend/src/utils/use_state_delayed.ts
```diff
@@ -0,0 +1,24 @@
+import { useState } from 'react';
+
+export function useStateDelayed(
+  initial: boolean,
+  config = { late: 25, prolong: 200 },
+): [{ now: boolean; prolonged: boolean; late: boolean }, (next: boolean) => void] {
```
pajowu, line 6: Can i have a short comment what this does please? The variable names don't really help me understand it

frontend/src/utils/use_blur.ts
pajowu: Also this should probably be called `useDebouncedOnClick` or something, right?

anuejn: Hm... not really, because it detects Clicks outside. maybe `useOnDebouncedOutsideClick`?

phlmn: My dear fried, I hope your foot is well

https://github.com/transcribee/transcribee/blob/9704d73cea481f43ecca9cb65cb23ed3ae725deb/frontend/src/editor/transcription_editor.tsx#L122

phlmn: > Also this should probably be called `useDebouncedOnClick` or something, right?

I vote for `useOnClickOutside`

PR #187, opened 2023-05-23
Batch load initial changes

rroohhh:
This seems to improve the initial load performance for me.
Going from ~1.6 seconds to 1.1 seconds.

### DIFF
--- a/frontend/src/editor/webvtt_export.tsx
+++ b/frontend/src/editor/webvtt_export.tsx
@@ -24,7 +24,7 @@ export function WebvttExportModal({
   const [format, setFormat] = useState('vtt' as SubtitleFormat);
 
   return (
-    <Modal {...props} onClose={onClose} label="Export as WebVTT">
+    <Modal {...props} onClose={onClose} label="Export as …">
       <form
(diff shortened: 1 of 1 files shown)


frontend/src/editor/webvtt_export.tsx
```diff
@@ -24,7 +24,7 @@ export function WebvttExportModal({
   const [format, setFormat] = useState('vtt' as SubtitleFormat);
 
   return (
-    <Modal {...props} onClose={onClose} label="Export as WebVTT">
+    <Modal {...props} onClose={onClose} label="Export as">
```
pajowu, line 27: Maybe just `Export`?

rroohhh: I am not that tied to this title, but `Export as` seems a common title for dialogs that let you choose the export format. 
Furthermore it acts as a title for the format dropdown, which I like.

anuejn: then I would at least add ...

pajowu: `…` is even better

PR #188, opened 2023-05-23
🔒️ Add check that users own document

pajowu:


backend/tests/conftest.py
```diff
@@ -48,6 +48,18 @@ def get_session_override():
...
+@pytest.fixture
+def logged_in_client_2(memory_session: Session, auth_token_2: str):
+    def get_session_override():
+        return memory_session
+
+    app.dependency_overrides[get_session] = get_session_override
+
+    client = TestClient(app, headers={"Authorization": f"Token {auth_token_2}"})
+    yield client
+    app.dependency_overrides.clear()
```
phlmn, line 60: Is it ok to clear other overrides here?

backend/tests/conftest.py
```diff
@@ -48,6 +48,18 @@ def get_session_override():
     app.dependency_overrides.clear()
 
 
+@pytest.fixture
+def logged_in_client_2(memory_session: Session, auth_token_2: str):
```
phlmn, line 52: Wow, this took me a while to get :D How about `logged_in_client_user_2` or something like that?

backend/tests/test_doc.py
```diff
@@ -44,10 +63,44 @@ def test_doc_delete(memory_session: Session, logged_in_client: TestClient):
...
+    req = getattr(client, method)(url.format(document_id=document))
+    assert 400 <= req.status_code < 500
+
+    req = getattr(logged_in_client_2, method)(url.format(document_id=document))
+    if need_specific_user:
+        assert 400 <= req.status_code < 500
+    else:
+        assert 200 <= req.status_code < 300
+
+    req = getattr(logged_in_client, method)(url.format(document_id=document))
```
phlmn, line 105: I would be happy about two comment like "try to acces with owning user" and "try to access with different user"

phlmn (changes requested):

rroohhh (approved):

PR #194, opened 2023-05-23
backend: use GZipMiddleware

rroohhh:


pajowu:
I don't see the advantage, IMHO this should be handled by the reverse proxy in front of our fastapi instance

PR #198, opened 2023-05-23
🚑️ Re-Add document to api task model

pajowu:
The document is required by the worker to download the media


rroohhh:
I would prefer something in this direction: https://paste.niemo.de/raw/ewejecidej.patch

rroohhh (approved):

PR #200, opened 2023-05-23
🚧 Allow document sharing

pajowu:


frontend/src/editor/share.tsx
```diff
@@ -0,0 +1,200 @@
...
+export function ShareModal({
+  onClose,
+  documentId,
+  ...props
+}: {
+  onClose: () => void;
+  documentId: string;
+} & Omit<ComponentProps<typeof Modal>, 'label'>) {
+  const [name, setName] = useState('Unnamed');
+  const [limitValidity, setLimitValidity] = useState(true);
```
anuejn, line 114: can I have the default be false here?

frontend/src/editor/share.tsx
```diff
@@ -0,0 +1,200 @@
...
+}
+export function ShareModal({
+  onClose,
+  documentId,
+  ...props
+}: {
+  onClose: () => void;
+  documentId: string;
+} & Omit<ComponentProps<typeof Modal>, 'label'>) {
+  const [name, setName] = useState('Unnamed');
```
anuejn, line 113: Can I have the current date included in the default here?

frontend/src/editor/share.tsx
```diff
@@ -0,0 +1,200 @@
...
+      {newToken && (
+        <div className="pt-4">
+          <h3 className="pb-1">Share Link</h3>
+          <div className={clsx('flex flex-row items-center ')}>
+            <Input className="flex-grow" readOnly value={getShareUrl(newToken)}></Input>
+            <IconButton
+              className="flex-grow-0"
+              icon={HiOutlineClipboardCopy}
+              label={'copy to clipboard'}
+              onClick={() => navigator.clipboard.writeText(getShareUrl(newToken))}
```
anuejn, line 188: Can I have an indication that the coppy worked?

anuejn (commented):
Works well in general, however I found a two annoyances:
* No error is displayed if one specifies an invalid / expired tokens
* No warning is displayed if one generates a token with an expiry date in the past
* No indication of being logged in is displayed. This might be a problem in the future but is okay for now?

frontend/src/editor/share.tsx
pajowu: I'd prefer it to be false, to nudge the user to set a limit (security by default). But I that is no strong opinion

pajowu:
Superseded by #312

PR #202, opened 2023-05-24
💅 add mobile optimized editor layout

anuejn:


phlmn:
Layout on large screens seems pretty dense to me
Screenshot 2023-05-24 at 13 39 19

The left side looks a bit odd on medium screens
Screenshot 2023-05-24 at 13 39 27

phlmn:
Dropdown receive a funny styling now when clicked?
image

phlmn:
Generellay I like the layout on mobile devices! :) Maybe a tad dense.

phlmn:
I guess this is WIP? Should we mark this as draft?

image

anuejn:
No, it actually is not - It is just badly tested ;). I would prefer to merge this soon-ish

pajowu:
With a long speaker name, the name is not always wrapped and the colored bar on the left side in the dropdown shrinks and eventually disappears. The bar also doesn't stretch if the speaker name is more than a line long

Screenshot 2023-05-28 at 21 04 58
Screenshot 2023-05-28 at 21 02 40

pajowu:
Other open points I still see:

- What happens if a speaker that spans more than one paragraph is changed? The labels are now only shown for the first paragraph in a block with the same speaker. I find it unintuitive that the dropdown of that label only changes the label of the first paragraph in that block and not the whole paragraph.
- Sometimes renaming a speaker only applies for newly rendered items or after reloading the page (I couldn't reproduce this though, maybe it already is fixed)
- Why is the name right-aligned, but the dropdown label left-aligned?
- Can I please have an indication of the currently selected speaker in the dropdown?

pajowu (changes requested):
See points to discuss

anuejn:
Okay, I think I incooperated all the feedback and am happy about a next round of review

pajowu (approved):

PR #224, opened 2023-05-25
anuejn/version

anuejn:
- ✨ display current transcribee version in frontend
- 🐞 fix useStateDelayed rapid toggling
- 🐞 fix tooltip bottom arrow
- 🐞 close tooltip with delay

### DIFF
--- a/frontend/src/common/version.tsx
+++ b/frontend/src/common/version.tsx
@@ -1,0 +1,74 @@
+import { Tooltip } from '../components/tooltip';
+import './version';
+// eslint-disable-next-line import/no-unresolved
+import version, { Commit } from 'virtual:git-version';
+import { IoMdOpen } from 'react-icons/io';
+import clsx from 'clsx';
+
+function formatDate(date: string): string {
+  return new Date(date).toLocaleString('de-DE', { dateStyle: 'short', timeStyle: 'short' });
+}
+
+export function CommitPopup({ commit, text }: { commit: Commit; text: string }) {
+  return (
+    <Tooltip
+      className="inline-block underline decoration-dashed"
+      placements="top"
+      tooltipText={
+        <>
+          Last commit{' '}
+          <a
+            href={`https://github.com/transcribee/transcribee/commit/${commit.hash}`}
+            target="_blank"
+            rel="noreferrer"
+            className="text-blue-600 underline dark:text-blue-500 hover:no-underline"
+            title="View commit on GitHub"
+          >
+            {commit.hash.substring(0, 10)} <IoMdOpen className="inline" />
+          </a>{' '}
+          on {formatDate(commit.date)}.
+        </>
+      }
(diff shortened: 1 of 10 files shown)


frontend/src/common/version.tsx
```diff
@@ -0,0 +1,59 @@
...
+        {versionText} + {diffShort}
+      </>
+    );
+  }
+
+  return (
+    <>
+      <div className="mb-10">
+        <div className="absolute bottom-0 h-10 text-xs left-1/2 -translate-x-1/2 opacity-50 hover:opacity-100 duration-400 transition-all">
+          Frontend build on {formatDate(version.date)}. {versionText}
```
pajowu, line 54:
```suggestion
          Frontend built on {formatDate(version.date)}. {versionText}
```

frontend/src/common/version.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <>
+      <Tooltip
+        className="inline-block underline decoration-dashed"
+        tooltipText={`until ${lastCommitOnMain.hash.substring(0, 10)} from ${formatDate(
+          lastCommitOnMain.date,
+        )}`}
+      >
+        {lastCommitOnMain.countSinceStart} commits on main
+      </Tooltip>
+    </>
```
pajowu, line 23: This could be a component

frontend/src/common/version.tsx
```diff
@@ -0,0 +1,59 @@
...
+  );
+
+  if (branch !== 'main') {
+    versionText = (
+      <>
+        {versionText} +{' '}
+        <Tooltip
+          className="inline-block underline decoration-dashed"
+          tooltipText={`until ${lastCommit.hash.substring(0, 10)} from ${formatDate(
+            lastCommit.date,
```
pajowu, line 33: I feel like this is not proper english?

pajowu (changes requested):
The version is currently not shown on the document editor page. Is this on purpose? I'd prefer to have it there as well

frontend/src/common/version.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export function Version() {
+  const { lastCommitOnMain, branch, lastCommit, diffShort } = version;
+
+  let versionText = (
+    <>
+      <Tooltip
+        className="inline-block underline decoration-dashed"
+        tooltipText={`until ${lastCommitOnMain.hash.substring(0, 10)} from ${formatDate(
```
rroohhh, line 17: Maybe a weird idea, but making this a link to github with the correct revision would be fun.
We could even implement a way to download the source that way.

frontend/src/common/version.tsx
anuejn: any suggestions?

pajowu (approved):

PR #228, opened 2023-05-26
Optimize  frontend

rroohhh:
This does two main things:
1. switch to a newer `slate-automerge-doc` version, which is using the unstable `Automerge` API
2. optimize the frontend performance by reducing rerenders
 
It introduces document hooks and remove some accumulated hacks.


rroohhh:
Needs https://github.com/transcribee/slate-automerge-doc/

Drastically improves performance by avoiding a lot of rerenders in addition to doing less automerge related work.

frontend/src/editor/automerge_websocket_editor.ts
```diff
@@ -82,14 +67,12 @@ export function useAutomergeWebsocketEditor(
   }, [editor]);
 
   useEffect(() => {
-    const onDocChange = editor.onDocChange;
-    editor.onDocChange = (newDoc) => {
+    editor.registerOnDocChange((newDoc: Document) => {
```
anuejn, line 70: I think we want to unregister the onDocChange again in cleanup?

frontend/src/editor/player.tsx
```diff
@@ -92,9 +92,8 @@ export function PlayerBar({ documentId, editor }: { documentId: string; editor:
   // skip to a timestamp if the user clicks on a word in the transcript. The corresponding event is
   // dispatched in transcription_editor.tsx
   useEvent<TextClickEvent>(TEXT_CLICK_EVENT, (e) => {
-    progressCallback();
-    if (e.detail.text.start) {
-      waveSurferRef.current?.seekTo(e.detail.text.start / waveSurferRef.current.getDuration());
+    if (e.detail.start != undefined) {
```
anuejn, line 95: good catch :D. but don't we still need to call progressCallback after that?

frontend/src/editor/speaker_dropdown.tsx
```diff
@@ -1,83 +1,78 @@
...
-  doc,
+  editor,
   onSpeakerSelected,
   onNewSpeaker,
   onClose,
   selected,
   ...props
 }: {
-  doc: Automerge.Doc<Document>;
+  editor: Editor;
```
anuejn, line 22: we can use useSlate / useSlateStatic here

frontend/src/editor/speaker_dropdown.tsx
```diff
@@ -159,94 +145,87 @@ function setSpeaker(editor: AutomergeEditor, path: number[], speaker: string | n
...
-        >
-          Unset Speaker
-        </DropdownItem>
-      </DropdownSection>
-    </Dropdown>
-  );
+SpeakerDropdownInner.displayName = 'SpeakerDropdownInner';
+
+export function SpeakerDropdown({ paragraph }: { paragraph: Paragraph }) {
+  const editor: Editor = useSlateStatic();
```
anuejn, line 229: does this really have the needed rerender guarantees? This component is not rerendered if a speaker name changes?!

frontend/src/editor/transcription_editor.tsx
```diff
@@ -40,7 +40,10 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
           <div
             contentEditable={false}
             className="w-16 mr-2 -ml-20 hidden 2xl:block text-slate-500 dark:text-neutral-400 font-mono"
-            onClick={() => window.dispatchEvent(new TextClickEvent(startAtom))}
+            onClick={useCallback(
```
anuejn, line 43: how is this better to just passing the callback directly?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -70,32 +75,54 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
...
+    conf,
+    children,
+    attributes,
+  }: {
+    start?: number;
+    conf?: number;
+    children: RenderLeafProps['children'];
+    attributes: RenderLeafProps['attributes'];
+  }): JSX.Element => {
+    const systemPrefersDark = useMediaQuery('(prefers-color-scheme: dark)');
```
anuejn, line 90: does it help performance to optimize this and don't do this once per word / batch it in useMediaQuery?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -70,32 +75,54 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
...
+      : systemPrefersDark
+      ? `hsl(0, 100%, ${Math.min(conf * 50 + 50, 100)}%)`
+      : `hsl(0, 100%, ${Math.min((1 - conf) * 100, 45)}%)`;
+
+    const classes = ['word'];
+    if (start !== undefined) {
+      classes.push(startTimeToClassName(start));
+    }
 
-  const systemPrefersDark = useMediaQuery('(prefers-color-scheme: dark)');
```
anuejn, line 79: whooops, good catch

frontend/src/editor/transcription_editor.tsx
```diff
@@ -70,32 +75,54 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
...
+    );
+  },
+);
 
-  const color = !leaf.conf
-    ? undefined
-    : systemPrefersDark
-    ? `hsl(0, 100%, ${Math.min(leaf.conf * 50 + 50, 100)}%)`
-    : `hsl(0, 100%, ${Math.min((1 - leaf.conf) * 100, 45)}%)`;
+Leaf.displayName = 'Leaf';
```
anuejn, line 118: what does this do? help debugging? Cann we do this in a less ugly-imperative way?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -70,32 +75,54 @@ function renderElement({ element, children, attributes }: RenderElementProps): J
...
 
+function renderLeaf(props: RenderLeafProps): JSX.Element {
```
anuejn, line 120: couldn't we just omit this function and pass Leaf directly? If not, cant we just return `<Leaf {...props />` and maybe inline this into the other component?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -123,20 +150,21 @@ export function TranscriptionEditor({
             /* the value is actually managed by the editor object */
           ]
         }
-        onChange={() => {
+        onChange={useCallback(() => {
```
anuejn, line 153: does this really help? did you try?

frontend/src/editor/transcription_editor.tsx
```diff
@@ -123,20 +150,21 @@ export function TranscriptionEditor({
...
           );
           if (hasChanged) {
             Transforms.setNodes(editor, { conf: 1.0 }, { match: (n) => 'conf' in n });
           }
-        }}
+        }, [])}
       >
         <Editable
-          renderElement={(props) => renderElement(props)}
+          renderElement={useCallback((props: RenderElementProps) => renderElement(props), [])}
```
anuejn, line 164: see above

frontend/src/editor/transcription_editor.tsx
```diff
@@ -123,20 +150,21 @@ export function TranscriptionEditor({
...
           renderLeaf={renderLeaf}
-          onClick={(e) => {
+          onClick={useCallback((e: React.MouseEvent) => {
```
anuejn, line 166: see above

frontend/src/editor/types.ts
```diff
@@ -20,8 +20,8 @@ export type Document = {
 
 // we fire this event when the user clicks on a word and we want the player to skip through it
 export const TEXT_CLICK_EVENT = 'textClick';
-export class TextClickEvent extends CustomEvent<{ text: Text }> {
-  constructor(text: Text) {
-    super(TEXT_CLICK_EVENT, { detail: { text } });
+export class TextClickEvent extends CustomEvent<{ start?: number }> {
```
anuejn, line 23: If we do this, I would prefer to rename this event to something like `SeekEvent` or `SeekToEvent`

frontend/src/utils/document.ts
```diff
@@ -0,0 +1,84 @@
+import { useEffect, useState, useCallback } from 'react';
+import { Editor } from 'slate';
+import { Document } from '../editor/types';
+
+export function getSpeakerName(
```
anuejn, line 5: good call to move this here!

frontend/src/utils/document.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (!speaker) {
+    return `Unknown Speaker`;
+  } else if (speaker in speaker_names) {
+    return speaker_names[speaker];
+  } else {
+    return `Unnamed Speaker ${speaker}`;
+  }
+}
+
+export function useDocument<T>(editor: Editor, callback: (doc: Document) => T) {
```
anuejn, line 18: can we call callback `selector` or `lens` ?

anuejn: also: why does this not use useSlate internally?

frontend/src/utils/document.ts
```diff
@@ -0,0 +1,84 @@
...
+        setSpeakerName(newSpeakerName);
+      },
+      [speakerID],
+    ),
+  );
+
+  return speakerName;
+}
+
+export function useSpeakerNames(editor: Editor): Record<string, string> {
```
anuejn, line 42: same

frontend/src/editor/speaker_colors.tsx
```diff
@@ -1,5 +1,6 @@
-import React, { useEffect, useState } from 'react';
+import React, { useMemo } from 'react';
 import { Editor } from 'slate';
+import { useSpeakerIDs } from '../utils/document';
 
 export const SpeakerColorsContext = React.createContext<Record<string, string>>({});
```
anuejn, line 5: I think at this point we could just remove this whole context hack and move the remaining code into `utils/document.ts`

anuejn: Also we might want to have a way to have memoized selectors onto the document at some point. maybe not now but lets create an issue?

pajowu:
This effectively breaks all existing documents. IMHO we need to either (a) bump the format version or (b) handle this properly in the frontend (and backend?)

frontend/src/editor/speaker_dropdown.tsx
```diff
@@ -1,83 +1,78 @@
...
+  const speakerNames = useSpeakerNames(editor);
 
-  if (speakerId === null && selected !== undefined && selected in speakerNames) {
+  if (speakerId === null && selected !== null && selected in speakerNames) {
     setSpeakerId(selected);
   } else if (speakerId === null && Object.keys(speakerNames).length > 0) {
     setSpeakerId(Object.keys(speakerNames)[0]);
   } else if (speakerId === null && Object.keys(speakerNames).length === 0) {
     setSpeakerId(NEW_SPEAKER_OPTION);
   }
```
pajowu, line 41: This breaks the "change speaker" modal: Since with `useSpeakerNames` `speakerNames` is `{}` initially and only filled with a value after a short delay, this will always go to the last branch of this if condition

frontend/src/editor/player.tsx
rroohhh: `progressCallback` is already called after seeking due to listening to the seek events, so I am not sure what calling it here aswell would really do?

frontend/src/editor/speaker_dropdown.tsx
rroohhh: It does, because editor is only shoved into the `useDocument` (or rather `useSpeakerName{,s}`.
But I'll change those hooks to make that clearer.

frontend/src/editor/transcription_editor.tsx
rroohhh: Well `eslint` fails without this. 
Not sure there is a different way, I did not find one.

frontend/src/editor/transcription_editor.tsx
rroohhh: This one definitely help a lot, because otherwise the `renderElement` function changes every time `TranscriptionEditor` is rendered, which in turn causes all of the elements to be rerendered.

frontend/src/utils/document.ts
rroohhh: It should use `useSlateStatic` internally, `useSlate` I am not convinced by, because using `onDocChange` we can avoid a lot of recalculations (because we only catch actual document changes, instead of changes to selection for example)

frontend/src/editor/transcription_editor.tsx
anuejn: if the component to render is cheap, this is not called for (see https://dmitripavlutin.com/react-usecallback/)

frontend/src/editor/transcription_editor.tsx
anuejn: if we dont use a lambda here, function equality should hold as-is

frontend/src/editor/speaker_dropdown.tsx
rroohhh: we cannot, the modals are not below slate in the react tree

frontend/src/editor/transcription_editor.tsx
rroohhh: A bit, about to 2ms

frontend/src/editor/speaker_colors.tsx
rroohhh: Well, the memoized selectors onto the document seem a lot like what context is doing to me?
I am not that fluent in react, is context not how you usually do thing?

anuejn: Yes but without the need to express it explicitly. for the time being I think we are good with just re-calculating it sometimes as it is neither very costly nor very often done

frontend/src/editor/transcription_editor.tsx
phlmn: This sets the name for the component in react dev tools. With normal functions and arrow functions it is able to set it automatically. With memo this has to be done by hand afaik

frontend/src/editor/types.ts
```diff
@@ -19,9 +19,9 @@ export type Document = {
 };
 
 // we fire this event when the user clicks on a word and we want the player to skip through it
-export const TEXT_CLICK_EVENT = 'textClick';
-export class TextClickEvent extends CustomEvent<{ text: Text }> {
-  constructor(text: Text) {
-    super(TEXT_CLICK_EVENT, { detail: { text } });
+export const SEEK_TO_EVENT = 'textClick';
+export class SeekToEvent extends CustomEvent<{ start?: number }> {
```
phlmn, line 23: It's a bit weird this is now named by what it should do, not by what has happened?

Nevertheless, if we keep this the string should be updated as well.

frontend/src/editor/transcription_editor.tsx
```diff
@@ -126,32 +140,49 @@ export function TranscriptionEditor({
...
-              Range.isCollapsed(selection) &&
-              e.target instanceof HTMLElement &&
-              e.target.isContentEditable
-            ) {
-              const [leaf] = editor.leaf(selection.anchor);
-              window.dispatchEvent(new TextClickEvent(leaf));
-            }
-          }}
-        />
+              // fire a text click event when selection is changed by clicking outside of a text node
```
phlmn, line 172: would be a 'seek to' event now

pajowu (approved):

PR #231, opened 2023-05-28
⌨️ Tap tab tab to skip

### LINKED ISSUE #216: Double tap <TAB> to go back 2 seconds

phlmn:
closes #216


frontend/src/editor/player.tsx
```diff
@@ -12,6 +12,12 @@ import { Editor } from 'slate';
 import { useButtonHoldRepeat } from '../utils/button_hooks';
 import { useLocalStorage } from '../utils/use_local_storage';
 
+const DOUBLE_TAP_THRESHOLD_MS = 250;
+const SKIP_BUTTON_SEC = 2;
+const SKIP_SHORTCUT_SEC = 3;
```
pajowu, line 17: Why do we use different values here?

phlmn: Because it feels better for me. Do you want to try it and see what values work for you the best?

pajowu: I'm fine with this, just wanted to know if there is a deeper reason

pajowu (approved):

PR #250, opened 2023-06-08
✨ Move task attempts to extra model, limit retries

pajowu:


backend/transcribee_backend/models/task.py
```diff
@@ -83,26 +102,65 @@ class Task(TaskBase, table=True):
...
+
+    started_at: Optional[datetime.datetime] = None
+    last_keepalive: Optional[datetime.datetime] = None
+    ended_at: Optional[datetime.datetime] = None
+
+    progress: Optional[float] = None
+
+    extra_data: Optional[dict] = Field(
+        sa_column=Column(JSON(), nullable=True), default=None
+    )
```
anuejn, line 137: maybe we want to add a number which attempt this is in the model?

backend/transcribee_backend/models/task.py
```diff
@@ -83,26 +102,65 @@ class Task(TaskBase, table=True):
...
+
+
+class TaskAttemptResponse(SQLModel):
+    progress: Optional[float]
```
anuejn, line 141: that is not a lot of information. Maybe we want to add at least the started time?

backend/transcribee_backend/routers/task.py
```diff
@@ -64,67 +78,101 @@ def claim_unassigned_task(
...
+    task_type: List[TaskType] = Query(),
+    now: datetime.datetime = Depends(now_tz_aware),
+) -> Optional[AssignedTaskResponse]:
+    while True:
+        task = get_ready_task(session, task_type)
+        if task is None:
+            session.commit()  # We might have marked tasks as failed
+            return
+        elif task.remaining_attempts <= 0:
+            mark_task_as_failed(session, task, now)
```
anuejn, line 109: A bit sad is, that we will not mark a task as failed if the task crashes the only connected worker and it does not come back?

backend/transcribee_backend/routers/task.py
```diff
@@ -64,67 +78,101 @@ def claim_unassigned_task(
...
-
-    task.last_keepalive = now_tz_aware()
+    task.current_attempt.last_keepalive = now_tz_aware()
     if keepalive_data.progress:
-        task.progress = keepalive_data.progress
+        task.current_attempt.progress = keepalive_data.progress
+        session.add(task.current_attempt)
     session.add(task)
     session.commit()
     return AssignedTaskResponse.from_orm(task)
```
anuejn, line 139: Do we actually ever do something if the keepalive does not come on our own in the backend? (or only if a worker polls us?)

anuejn (commented):
Looks good, just a few minor questions :)

backend/transcribee_backend/models/task.py
pajowu: This is only used for communication with the frontend (this is not a database model). Since we don't currently use this information in the frontend, I'd rather keep it as is and change it once the frontend needs this information

worker/transcribee_worker/config.py
```diff
@@ -24,6 +24,8 @@ class Settings(BaseSettings):
         },
     }
 
+    KEEPALIVE_INTERVAL: float = 0.5
```
rroohhh, line 27: Not that important, but this seems like a very short interval? 
I would have thought we might use something on the order of 5 seconds.

pajowu: Many tasks provide much finer progress updates and showing this in the UI quickly make the progress feel faster (at least that's what I found during development). Additionally we can change this using the env variable for every worker

backend/transcribee_backend/main.py
```diff
@@ -30,3 +35,10 @@ async def root():
 
 
 app.get("/media/{file}")(serve_media)
+
+
+@app.on_event("startup")
+async def setup_repeating_tasks():
```
anuejn, line 41: again: naming

anuejn (approved):
Looks good to me, just one nitpick

PR #256, opened 2023-06-13
🚑️ Run add_document_media_file in executor

pajowu:
Otherwise the keepalive task wouldn't be run


pajowu:
This is a workaround for a crash that workers currently experience: During the upload `keepalive_task` is not executed. If the upload takes too long, the task is marked as failed by the backend.

We could consider switching to async http in general, but I'd like to merge this workaround quickly first

anuejn (approved):

PR #267, opened 2023-06-20
🎨 Add ruff & pyright

pajowu:


backend/transcribee_backend/routers/document.py
```diff
@@ -253,10 +253,11 @@ def add_media_file(
...
-    if task is None:
-        raise HTTPException(status_code=403)
-
-    if task.current_attempt.assigned_worker != authorized_worker:
+    if (
+        task is None
+        or task.current_attempt is None
+        or task.current_attempt.assigned_worker != authorized_worker
+    ):
         raise HTTPException(status_code=403)
```
phlmn, line 261: why do we answer with 'forbidden' here?

backend/transcribee_backend/routers/document.py
```diff
@@ -295,10 +296,11 @@ def set_duration(
...
```
phlmn, line 304: why do we answer with 'forbidden' here?

backend/transcribee_backend/routers/task.py
```diff
@@ -103,6 +104,10 @@ def keepalive(
     session: Session = Depends(get_session),
     task: Task = Depends(get_authorized_task),
 ) -> Optional[AssignedTaskResponse]:
+    # mostly to please the type checker, get_authorized_task already ensures
+    # that the task has a current attempt
+    if task.current_attempt is None:
+        raise HTTPException(status_code=403)
```
phlmn, line 110: sounds like an server error then?

backend/transcribee_backend/routers/document.py
pajowu: What would you propose? 404?

backend/transcribee_backend/routers/document.py
pajowu: What would you propose? 404?

backend/transcribee_backend/routers/document.py
pajowu: I decided for a 403 here, because this is a document endpoint and the document exists, so a 404 seems wrong. The problem is that the document may only be changed if a matching task exists that is assigned to this worker, which this checks for. If the task does not exist, the worker is not allowed to call this api, but another worker might. So I think 403 is the correct status code here

rroohhh:
Are the migrations not automatically generated? Imo auto-generated files should not be type-checked / checked for formatting

pajowu:
The migrations are only partially auto-generated. I manually modified some of them, so I'd like to keep the typechecking & formatting for them

backend/transcribee_backend/routers/document.py
phlmn: It depends... If the user is not allowed to see the task a 404 should be the correct response IMO.

backend/transcribee_backend/routers/document.py
phlmn: Yes

rroohhh (approved):

PR #269, opened 2023-06-22
✨ Allow changing password

### LINKED ISSUE #171: Change Password
We should allow users to change their password. This needs an api call for it as well as ui

pajowu:
Fixes #171


anuejn (approved):
Looks good to me :)

PR #284, opened 2023-06-27
🐛 fix document loading in firefox

anuejn:


frontend/src/editor/automerge_websocket_editor.ts
```diff
@@ -62,17 +62,31 @@ export function useAutomergeWebsocketEditor(
...
         console.log('Received new document');
         doc = Automerge.load(msg);
-        console.log('Loaded new document');
+        if (!initialSync) {
+          const migratedDoc = migrateDocument(doc as Automerge.Doc<Document>);
+          sendDocChange(migratedDoc);
+
+          editor.setDoc(migratedDoc);
+          const setDocTime = Date.now();
+          console.info(`setDoc in ${(setDocTime - Date.now()) / 1000} s`);
```
rroohhh, line 87: This looks like it times exactly nothing

frontend/src/editor/automerge_websocket_editor.ts
```diff
@@ -62,17 +62,31 @@ export function useAutomergeWebsocketEditor(
...
         editor.setDoc(migratedDoc);
+        const setDocTime = Date.now();
+        console.info(`setDoc in ${(setDocTime - syncCompleteTime) / 1000} s`);
+
         onInitialSyncComplete();
       } else if (msg_type === MessageSyncType.FullDoc) {
         console.log('Received new document');
         doc = Automerge.load(msg);
-        console.log('Loaded new document');
+        if (!initialSync) {
```
rroohhh, line 81: Why is this gated on `!initialSync`?

rroohhh (changes requested):
IIRC this fixes document loading whenever the sync messages occur out of order. 
With websocket this should be impossible, so this feels more like a hacky workaround. 

Were you able to determine any reason why these messages might be sent out of order?

frontend/src/editor/automerge_websocket_editor.ts
anuejn: because otherwise it will be batch-applied with all the other changes on sync complete

frontend/src/editor/automerge_websocket_editor.ts
anuejn: whoops, true

anuejn:
Sadly I was not able to determine the reason, why the messages appear OoO.
However, these fixes also make sense if for some reason we get a new document for different reasons

frontend/src/editor/automerge_websocket_editor.ts
```diff
@@ -62,17 +62,31 @@ export function useAutomergeWebsocketEditor(
           editor.setDoc(newDoc);
         }
       } else if (msg_type === MessageSyncType.ChangeBacklogComplete) {
+        console.info('backlog complete');
```
rroohhh, line 65: I feel like this "batch load + migrate" should be factored out into a function or lambda and be used by the `ChangeBacklogComplete` and the `FullDoc` path

rroohhh (approved):

PR #285, opened 2023-06-27
Issie-283: logout fix

voronov007:
Remove UserToken from DB when logout.

Fix Issue-#283

PR #288, opened 2023-07-04
Remove auth tokens from db on logout & after end of validity

### LINKED ISSUE #283: Remove auth tokens from db on logout & after end of validity
Currently the auth tokens just linger in the database forever. This should be changed in two ways:

1. If a user clicks "logout", an api call should be made that removes this login token from the database
2. A periodic task should be added that removes all expired auth tokens from the database

pajowu:
Based on #285 , Fixes #283

### DIFF
--- a/frontend/src/api/user.ts
+++ b/frontend/src/api/user.ts
@@ -11,8 +11,8 @@ export const changePassword = fetcher
   .path('/api/v1/users/change_password/')
   .method('post', 'application/json')
   .create();
-
 export const getMe = fetcher.path('/api/v1/users/me/').method('get').create();
+export const logout = fetcher.path('/api/v1/users/logout/').method('post').create();
 
 const useGetMeWithRetry = makeSwrHook('getMe', getMe);
 
--- a/frontend/src/common/top_bar.tsx
+++ b/frontend/src/common/top_bar.tsx
@@ -1,6 +1,6 @@
 import { useLocation } from 'wouter';
 import { storeAuthToken } from '../api';
-import { useGetMe } from '../api/user';
+import { useGetMe, logout } from '../api/user';
 import { primitiveWithClassname } from '../styled';
 import { BiUser } from 'react-icons/bi';
 import { IconButton, PrimaryButton, SecondaryButton } from '../components/button';
@@ -38,7 +38,12 @@ export function MeMenu() {
           Change Password
         </SecondaryButton>
         <PrimaryButton
-          onClick={() => {
+          onClick={async () => {
+            try {
+              await logout({});
+            } catch (e) {
+              /* empty */
+            }
             storeAuthToken(undefined);
             mutate();
             navigate('/');
--- a/backend/openapi-schema.yml
+++ b/backend/openapi-schema.yml
@@ -953,6 +953,29 @@ paths:
                 $ref: '#/components/schemas/HTTPValidationError'
           description: Validation Error
       summary: Login
+  /api/v1/users/logout/:
+    post:
+      operationId: logout_api_v1_users_logout__post
+      parameters:
+      - in: header
+        name: authorization
+        required: true
+        schema:
+          title: Authorization
+          type: string
+      responses:
+        '200':
+          content:
+            application/json:
+              schema: {}
+          description: Successful Response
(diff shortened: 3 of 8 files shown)


pajowu:
@voronov007 I added the commits from our Pull Request to this new branch and pull request. I'll add some more small changes and squash the commits later today and then it could be merged from my side (still needs review from the other though)

frontend/src/api/user.ts
```diff
@@ -11,8 +11,8 @@ export const changePassword = fetcher
   .path('/api/v1/users/change_password/')
   .method('post', 'application/json')
   .create();
-
 export const getMe = fetcher.path('/api/v1/users/me/').method('get').create();
+export const logout = fetcher.path('/api/v1/users/logout/').method('post').create();
```
voronov007, line 15: This method needs to be a `GET` method, because it requires a user token, which is sent in `GET` request. Otherwise, the `POST` method needs to have a token in it (so needs to be updated on Frontend side)

pajowu: The auth token is sent for all request methods: https://github.com/bugbakery/transcribee/blob/28752bc944554b4b815186fd9cd0a1622e0ad695/frontend/src/api.ts#L13

voronov007: then it is not recognized as `token: UserToken = Depends(get_user_token)`. I think in such a case argument needs to be `session: Session = Depends(get_session),`. On the local machine I checked that with GET request solution worked, while POST request - didn't work.

frontend/src/common/top_bar.tsx
```diff
@@ -38,7 +38,12 @@ export function MeMenu() {
...
         <PrimaryButton
-          onClick={() => {
+          onClick={async () => {
+            try {
+              await logout({});
+            } catch (e) {
+              /* empty */
+            }
             storeAuthToken(undefined);
             mutate();
```
voronov007, line 48: I think this call is redundant. It tries to get info about the user and gets 404 that user is not found (because the token was erased even with `storeAuthToken(undefined)`)

frontend/src/api/user.ts
pajowu: `get_session` is not related to the user auth. It get the database session, i.e. the connection to the database. `get_user_token` is the correct dependency here, which works with `POST` requests. I tested this PR, it works the way it is currently implemented. We also use `POST` requests with `get_user_token` in many other places, for example when changing the password https://github.com/bugbakery/transcribee/blob/28752bc944554b4b815186fd9cd0a1622e0ad695/backend/transcribee_backend/routers/user.py#L63

frontend/src/common/top_bar.tsx
pajowu: Yes, it is probably reduntant in the current setup, especially since we call `window.location.reload();` a few lines later: `mutate` marks the swr cache as expired and refetches the data[^mutate], but we currently do not keep the data in the browser cache across reloads. I would still keep it, as it seems to me to be cleaner to mark the data as invalid. Especially if we ever implement some form of data sharing across tabs/reloads, this might be needed.

[^mutate]: https://swr.vercel.app/docs/mutation#revalidation

frontend/src/api/user.ts
voronov007: ok, cool. A double-check was necessary.

voronov007:
MR looks good to me.

@pajowu just FYI: during a job search I got a task (from an employer) to contribute to some Open Source project with Python backend. I found this project and this issue by the `good first issue` label. I hope that the changes that I provided as a part of my contribution are helpful for the project. 

A bit of feedback, comments, and recommendations:
- I would recommend separate BE and FE. Following the setup documentation, it takes a few minutes to start the service (without building)
- I was able to run the project only from Docker, but bind-mount option didn't work and I had to stop&run docker once I made some changes on FE or BE which made development very slow
- Running BE unit tests should be possible just with virtualenv. Frankly speaking, I was not able to run existing unit tests on a local machine (Ubuntu) as I got errors. I didn't try to run unit tests from the Docker.

I wish you good luck with the project. All the best!

phlmn (approved):

PR #292, opened 2023-07-06
anuejn/performance fiddling

anuejn:
- init logout fixl
- updated openapi docs; lint fix;
- add periodic task; updated docker updates;
- fixed logout;
- lint fix;
- change logout http method to post
- cleanup


anuejn:
superseeded by #311

PR #294, opened 2023-07-10
✨ Debug import/export

pajowu:


doc/development_setup.md
```diff
@@ -44,6 +44,21 @@ the following steps:
 > docker run -it -p 5173:5173 --mount type=bind,source="$(pwd)",target=/app transcribee:latest
 > ```
 
+## Debug Mode
+
+Some features are only available in debug mode.
+To enable this mode, set the `DEBUG_MODE` environment variable to `true`.
+
+```
+DEBUG_MODE=true ./packaging/dev.sh
```
anuejn, line 53: do we maybe enable this as a default in dev.sh?

frontend/src/pages/user_home.tsx
```diff
@@ -1,16 +1,25 @@
...
+import { showModal, Modal } from '../components/modal';
+import { ComponentProps, useRef, useState } from 'react';
+import { DropdownItem, DropdownSection, IconDropdown } from '../components/dropdown';
+import { useDebugMode } from '../debugMode';
+
+type DocumentInfo = ReturnType<typeof useListDocuments>['data'][0];
 
 export function UserHomePage() {
   const { data, mutate } = useListDocuments({});
+  const debugMode = useDebugMode();
```
anuejn, line 22: this can be in the DocumentCard.
also: it is kinda wiered, that this is a different DebugMode than in server? Do we really need both? can we not just enable the server functionality always?

anuejn (commented):
Some additional observations:
* I would whish for the possibility to create empty documents
* When one imports documents while worker jobs are running, the document is broken
* There is some confusion around the different debugModes going on

frontend/src/pages/user_home.tsx
pajowu: Oh, good catch. This *should be* the backend debug mode. VSCode auto import resolution tricked me

frontend/src/editor/export/export_modal.tsx
```diff
@@ -0,0 +1,95 @@
...
+        onSubmit={async (e) => {
+          e.preventDefault();
+          await onExportRef.current(editor);
+          onClose();
+        }}
+      >
+        <Select
+          value={exportTypes.indexOf(exportMode)}
+          onChange={(e) => {
+            setExportMode(exportTypes[parseInt(e.target.value)]);
```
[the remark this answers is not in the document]
pajowu: What do you mean with "does not work for me"

pajowu:
Regarding the comments:

> I would whish for the possibility to create empty documents

I don't have an opinion on this, but would prefer to do this in a later PR.

> When one imports documents while worker jobs are running, the document is broken

Good point. Do you have a solution what to do against that? Since this is only a debug feature not intended for production use, I would be okay with just leaving this as is. You can fix the document by re-importing the document state once the worker jobs are done.

>There is some confusion around the different debugModes going on

This should be fixed now, good catch

frontend/src/pages/user_home.tsx
pajowu: Regarding always enabling the server functionality: I'm strongly opposed to this, as this is a very internal tool that should not be possible during production use. And if we don't show the option in the frontend, we should disable it in the backend too.

PR #295, opened 2023-07-17
💚 Build frontend in CI

pajowu:


anuejn (approved):
"We'll burn that bridge when we get there"

PR #302, opened 2023-07-19
🐛 Use automerge.Text in speaker id if only 1 para

pajowu:
This fixes the frontend showing `Unnamed Speaker [object Object]` in documents with only one paragraph


phlmn (approved):
lol

PR #312, opened 2023-07-27
✨ Allow document sharing

pajowu:


frontend/src/app.tsx
```diff
@@ -31,10 +32,21 @@ export function App() {
...
-            <Route path="/document/:documentId" component={DocumentPage} />
-            <Route component={PageNotFoundPage} />
           </>
         )}
+
+        {(isLoggedIn || hasShareToken) && (
+          <Route path="/document/:documentId" component={DocumentPage} />
+        )}
+        {hasShareToken && (
+          <>
```
anuejn, line 42: useless fragmemt

frontend/src/app.tsx
```diff
@@ -31,10 +32,21 @@ export function App() {
...
+            <Route>
+              <Redirect to="/login" />
```
anuejn, line 44: I whish for a comment explaining this

frontend/src/components/modal.tsx
```diff
@@ -97,12 +101,22 @@ export function Modal({
...
             {children}
-          </Dialog>
+          </DialogComponent>
         </div>
       </div>
     </div>
   );
 }
+
+export function DoubleWidthModal(
```
anuejn, line 114: Could I have something better here?
(variable width dialogs would be very nice)

anuejn (changes requested):
There are still some possible improvements I have found:
- The "no document popup content" popup is now stuck (not triggered by hover; doesnt even go away when there is document content) on new documents
- The popup breaks on Mobile
- I would like to have a read-only indication for existing tokens
- The label of the name input field is centered, that is not in accordance to our other UI-language
- I would prefer the "limit validity" to be something like "expiry date"
- I am not convinced of the centered table of the token list
- The coppied popup feels bad

anuejn (approved):

PR #313, opened 2023-07-27
 🚸 Make automatic speaker names 1-indexed

pajowu:
- 🚸 Make automatic speaker names 1-indexed
- 🐛 Fix typings for worker callback


anuejn (approved):
LGTM

PR #317, opened 2023-07-31
🐛 Fix speaker buttons showing over speakerdropdown

### LINKED ISSUE #316: Speaker buttons on hover shown above speaker dropdown

pajowu:
To be honest, I don't understand *why* this fixes it, but it does?

Fixes #316

### DIFF
--- a/frontend/src/components/dropdown.tsx
+++ b/frontend/src/components/dropdown.tsx
@@ -96,6 +96,9 @@ export function Dropdown({
           'hover:bg-gray-200 dark:hover:bg-neutral-700',
           show.now && 'bg-gray-200 dark:bg-neutral-700 !opacity-100',
           'px-2 py-1',
+          'relative',
+          'hover:z-10',
+          show.prolonged && 'z-50',
           buttonClassName,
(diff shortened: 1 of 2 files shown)


phlmn:
I think this re-introduces problems when the dropdown is higher than the paragraph... Then it is rendered behind the next speaker name. Maybe something like `hover:z-10` does the job?

pajowu:
I can't replicate that (if that is what you meant?)

Screenshot 2023-07-31 at 17 42 17

Screenshot 2023-07-31 at 17 42 21

phlmn:
Sorry, I meant the button to open the dropdown: 
image

pajowu:
Ah, now I understand. `hover:z-10` seems to be the solution, which I now added instead

phlmn:
It's even more cursed... I pushed a fix, but I don't like the code...

anuejn:
so... can we merge this?

pajowu:
IMO yes, just needs an approving review

anuejn (approved):

PR #325, opened 2023-08-02
⚡️ Resume transcription tasks

pajowu:


worker/transcribee_worker/worker.py
```diff
@@ -166,14 +176,21 @@ async def transcribe(
         audio = self.load_document_audio(task.document)
 
         async with self.api_client.document(task.document.id) as doc:
+
             async with doc.transaction("Reset Document") as d:
```
anuejn, line 180: we might want to use a different / no name for this transaction
(or dont execute this transaction and hoist the if statement one layer up)

anuejn: Automerge does not make a transaction if it is empty

pajowu: Automerge doesn't generate a change if no change is happening (which should always be the case)

anuejn (approved):

PR #327, opened 2023-08-03
⚡️ Only render children if in view 

pajowu:
Rendering just the text (so that searching and scrolling still kindof work) leads to better initial loading performance


anuejn (approved):
Very smart Idea & saves us a lot of dom elements :))

PR #336, opened 2023-08-07
🐬 re-introduce complex version

anuejn:
- ♻️ Reflect change of github orga name to bugbakery
- 🐬 re-introduce complex version


pajowu:
The complex version was added to the about page in #343

PR #337, opened 2023-08-07
⚡️ Improve error handling performance

pajowu:
We don't need a full rerender, we just clear the selection

https://github.com/ianstormtaylor/slate/pull/5407#issuecomment-1662449620


anuejn (approved):
Nice one :)

PR #345, opened 2023-09-01
fix loosing cursor position on new paragraph

anuejn:
this makes performance a bit worse but usability a lot better


pajowu (approved):

phlmn:
Wow.... the loading spinner is really bad UX. It does produce a white page even in dark mode and you cannot press the back button to cancel loading a large document anymore.

PR #348, opened 2023-09-03
🐛 Fix document import for documents with >1000 changes

pajowu:


anuejn (commented):
Why is it broken with >1000 changes in the first place? Why can't we process the changes on the server. Actually that seems like a somewhat more natural way to do it than sending the changes one by one?

pajowu:
Starlette only allows 1000 files “for security reasons” -> once we upload more than 1000, we get an http 422 error

anuejn (approved):
Okay, fair :)

PR #360, opened 2023-09-17
Add debian+nix dev container

moeffju:
This commit adds a dev container configuration file for a debian bullseye + nix container. Supporting IDEs, such as VS Code, should offer to open the cloned repository in the dev container, where users can then follow the [minimal setup](https://github.com/bugbakery/transcribee/blob/main/doc/development_setup.md#minimal-setup-nix) steps to quickly have a running dev instance without fiddling with their linux distro or nix locally.

It's a little bit more comfortable for VS Code users than the minimal docker setup.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,4 @@
+FROM ghcr.io/xtruder/nix-devcontainer:v1
+
+# cache /nix
+VOLUME /nix
--- a/.devcontainer/compose.yml
+++ b/.devcontainer/compose.yml
@@ -1,0 +1,16 @@
+version: '3'
+services:
+  dev:
+    build:
+      dockerfile: ./Dockerfile
+      args:
+        USER_UID: ${UID:-1000}
+        USER_GID: ${GID:-1000}
+    volumes:
+      - ..:/workspace:cached
+      - nix:/nix
+    security_opt:
+      - label:disable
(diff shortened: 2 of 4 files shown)


pajowu:
Nice, this also enables github codespaces, right?

moeffju:
Yeah, I believe that should also enable codespaces!

pajowu:
I think not running nix-shell or dev.sh is fine (at least for now), but could you please add some documentation on how to use this to doc/development_setup.md

pajowu:
Also sadly this doesn't seem to work in codespaces right now due to some bug https://github.com/NixOS/nix/issues/6680

moeffju:
I tried running a codespace on my fork and it seemed to work, but I'll try again.
Pushing commits for auto-building the nix env and running packaging/dev on first container run too, and will add documentation (although there isn't a lot to say, conforming IDEs should just notify the user).

moeffju:
Another thing that might be useful is to mount /nix into a volume so it persists between rebuilds. Currently if you rebuild the container, the nix cache is blown away as well, which is a bit annoying.

pajowu:
At least in code spaces the image does not build with the following error message:
```
The Compose file is invalid because:
Service dev has neither an image nor a build context specified. At least one must be provided.
```
Screenshot 2023-09-18 at 11 23 25

pajowu (approved):

PR #361, opened 2023-09-17
Add /.pnpm-store to gitignore

moeffju:
Following the instructions in the dev docs, `/.pnpm-store` gets a ton of files installed. Git dutifully and needlessly tries to scan for changes there, which is slow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .direnv
+/.pnpm-store


phlmn:
That's funny, normally `.pnpm-store` should be located in your home directory.

moeffju:
That may be an artifact of the dev container setup which maps the workspace folder and changes a bunch of env variables. I can look into overriding that later.

pajowu (approved):

PR #371, opened 2023-10-08
logged out redirect

anuejn:
this can be used to redirect to about.transcribee.net

### DIFF
--- a/backend/transcribee_backend/config.py
+++ b/backend/transcribee_backend/config.py
@@ -15,6 +15,7 @@ class Settings(BaseSettings):
     task_attempt_limit = 5
 
     media_url_base = "http://localhost:8000/"
+    logged_out_redirect_url: None | str = None
 
     model_config_path: Path = Path("data/models.json")
     pages_dir: Path = Path("data/pages/")
@@ -28,6 +29,7 @@ class ModelConfig(BaseModel):
 
 class PublicConfig(BaseModel):
     models: Dict[str, ModelConfig]
+    logged_out_redirect_url: str | None
 
 
 class ShortPageConfig(BaseModel):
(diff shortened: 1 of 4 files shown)


backend/transcribee_backend/config.py
```diff
@@ -69,7 +71,10 @@ def get_short_page_config() -> Dict[str, ShortPageConfig]:
 
 
 def get_public_config():
-    return PublicConfig(models=get_model_config())
+    return PublicConfig(
+        models=get_model_config(),
+        logged_out_redirect_url=environ.get("TRANSCRIBEE_LOGGED_OUT_REDIRECT_URL"),
```
pajowu, line 76: Please pull this from the `Setttings` instance, which gives more flexible ways of setting this (for example via a .env file)

frontend/src/app.tsx
```diff
@@ -47,6 +48,26 @@ export function LoggedInRoute<T extends DefaultParams = DefaultParams>({
...
+}: RouteProps<T>) {
+  const [_, navigate] = useLocation();
+  const { isLoggedIn, isLoading } = useAuthData();
+  const { data: config, isLoading: configLoading } = useGetConfig({});
+  if (isLoading) {
+    return <Route component={LoadingPage} />;
+  }
+  if (!isLoggedIn && !configLoading) {
+    if (config.logged_out_redirect_url) {
+      window.location.replace(config.logged_out_redirect_url);
```
pajowu, line 62: Souldnt this be assign, so the user can go back?

pajowu (changes requested):

frontend/src/app.tsx
anuejn: If we use assign and the user gos back, they are landing in an infinite loop. That would be way worse ux imo

backend/transcribee_backend/config.py
anuejn: Aaah I did not understand that, sounds good :)

pajowu (approved):

PR #384, opened 2023-11-16
Add worker api

pajowu:
- ✨ Add api to create workers
- ✨ Add endpoint to list workers


frontend/src/openapi-schema.ts
```diff
@@ -445,6 +458,33 @@ export interface components {
...
+       * Format: date-time
+       */
+      last_seen?: string;
+      /** Name */
+      name: string;
+      /** Token */
+      token: string;
+    };
+    /** WorkerBase */
+    WorkerBase: {
```
phlmn, line 479: `id` should probably be included here

pajowu: No, only a `transcribee_backend.models.worker.Worker` has a id. The worker base doesn't

phlmn: In know, but would be good to have in the response

pajowu: Why? The worker id is only used internally in the backend. The only thing needed via the api is the token / name

phlmn:
Can I have worker deregistration/deletion? 🥺👉👈

phlmn (approved):

PR #389, opened 2023-11-17
✨ Add metrics endpoint

pajowu:


backend/transcribee_backend/metrics.py
```diff
@@ -0,0 +1,128 @@
...
+from transcribee_backend.models.worker import Worker
+
+
+class Metric:
+    @abstractmethod
+    def refresh(self, session: Session):
+        pass
+
+
+class TasksInState(Metric):
```
anuejn, line 26: I would rather like to have this as TaskChanis or TaskGraphs in state. Otherwise this number is not really saying much (Transcode is much faster, yadayada)

backend/transcribee_backend/metrics.py
```diff
@@ -0,0 +1,128 @@
...
+        self.collector = Gauge("documents", "Documents")
+
+    def refresh(self, session: Session):
+        (result,) = session.query(func.count(Document.id)).one()
+        self.collector.set(result)
+
+
+class Queue(Metric):
+    def __init__(self):
+        self.collector = Gauge("queue", "Queue length in seconds")
```
anuejn, line 79: Same here. I would like to have every task chain only counted once

anuejn (changes requested):

backend/transcribee_backend/metrics.py
anuejn: Or maybe we introduce some unit here that is not seconds (maybe something like estimated compute seconds on some specific platform? the mac mini?)

backend/transcribee_backend/metrics.py
anuejn: I would prefer some other task parameters (that influence needed computing time) as labels then too

backend/transcribee_backend/metrics.py
pajowu: Yes and no. Doing a better calculation that takes into account the real time factor of different machines etc would be a nice addition, but I'd rather do that in a seperate PR, as this has some more open questions to me (and probably needs some database changes as well)

anuejn: Then for the time being only count transcription tasks?

PR #392, opened 2023-11-18
⚡️ Increase performance of list_documents by eager loading

pajowu:
On my device with 12 documents, this brings down the load time from ~90ms to ~18ms and the number of queries performed for one request to `/api/v1/documents/` from 194 to 7


anuejn (approved):
LGTM

PR #394, opened 2023-11-26
feat(frontend) - Export to Podlove Publisher

dnkbln:
Hi,

I add a new export option. The user can export the final transcript to a Podlove Publisher instance.

### DIFF
--- a/frontend/src/editor/export/webvtt.tsx
+++ b/frontend/src/editor/export/webvtt.tsx
@@ -152,9 +152,16 @@ export function WebVttExportBody({ onClose, outputNameBase, editor }: ExportProp
           disabled={!limitLineLength}
         />
       </FormControl>
-      {!canExport.canGenerate && canExport.reason && (
+      {((!canExport.canGenerate && canExport.reason) ||
+        (format === 'podlove' && !podloveExportPossible)) && (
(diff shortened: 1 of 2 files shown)


pajowu (changes requested):
Thank you so much for this PR. We're also working on adding an API that exposes the SRT/WebVTT file of a document, so maybe another option would be to wait for this API and then write an external script that syncs the two. But I'm not strictly opposed against adding this to transcribee directly

frontend/src/editor/export/webvtt.tsx
```diff
@@ -22,13 +29,14 @@ export function WebVttExportBody({ onClose, outputNameBase, editor }: ExportProp
...
           onChange={(e) => {
-            if (e.target.value === 'srt' || e.target.value === 'vtt') {
+            if (e.target.value === 'srt' || e.target.value === 'vtt' || e.target.value === 'podlove') {
               setFormat(e.target.value);
             }
           }}
         >
           <option value="vtt">WebVTT</option>
           <option value="srt">SRT</option>
+          <option value="podlove">Podlove</option>
```
pajowu, line 39: Maybe this could be renamed to something like "Upload to Podlove Publisher"

dnkbln:
I integrate your feedback.

pajowu (approved):

PR #402, opened 2023-12-04
✨ Add task queue info endpoint

phlmn:
Needed for autoscaling

### DIFF
--- a/backend/transcribee_backend/models/task.py
+++ b/backend/transcribee_backend/models/task.py
@@ -1,6 +1,7 @@
 import datetime
 import enum
 import uuid
+from dataclasses import dataclass
 from typing import Any, Dict, List, Literal, Optional
 
 from sqlmodel import JSON, Column, Field, ForeignKey, Relationship, SQLModel, col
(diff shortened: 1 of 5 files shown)


backend/transcribee_backend/models/task.py
```diff
@@ -64,6 +65,24 @@ class Task(TaskBase, table=True):
     state: TaskState = TaskState.NEW
     state_changed_at: datetime.datetime = Field(default_factory=now_tz_aware)
 
+    initial_cost: (float | None) = None
+
+    def guess_initial_cost(self, media_duration_sec: float) -> float | None:
```
pajowu, line 70: Why not calculate this as-needed and make this a property function?
```suggestion
    @property
    def initial_cost(self, media_duration_sec: float) -> float | None:
```

backend/transcribee_backend/routers/document.py
```diff
@@ -521,6 +521,11 @@ def set_duration(
 ) -> ApiDocument:
     doc = task.document
     doc.duration = body.duration
+
+    # update task costs based on duration
+    for task in doc.tasks:
+        task.initial_cost = task.guess_initial_cost(doc.duration)
+
```
pajowu, line 528: Can be removed if initial_cost is a property

backend/transcribee_backend/models/task.py
phlmn: We can do that, but then we have to query the document each time

phlmn: done

PR #406, opened 2023-12-09
✨ Allow user to disable video preview

pajowu:


frontend/src/editor/player.tsx
```diff
@@ -42,38 +43,53 @@ export function PlayerBar({
...
       data?.media_files.filter((media) => !media.tags.includes('original')) || [];
 
+    // but if the original is all we have, better play this than nothing at all
+    if (
+      relevantMediaFiles.length == 0 &&
+      data?.media_files !== undefined &&
+      data?.media_files.length > 0
+    ) {
+      relevantMediaFiles = data.media_files;
+    }
```
phlmn, line 58: then we could just adjust the sort function to append the original as last item, right?

phlmn: Also we then should prevent the player from loading the (large) original file in background while the loading bee is displayed and the reencode task did not finish yet.

phlmn (changes requested):

frontend/src/editor/player.tsx
pajowu: done (both)

frontend/src/editor/player.tsx
```diff
@@ -40,40 +63,27 @@ export function PlayerBar({
...
-        type: media.content_type,
-      };
-    });
-
-    return {
-      sources: sortMediaFiles(mappedFiles),
-      hasVideo: videoFiles.length > 0,
-    };
-  }, [data?.media_files]);
+  const [_showVideo, setShowVideo] = useState(true);
```
phlmn, line 73: Let's please name the state variable consistent with the setter :D

phlmn: Nevertheless approved already :)

phlmn (approved):

phlmn (changes requested):
Hm... just tested this and I get two competing audio streams playing at the same time, when disabling the video...

phlmn:
Also the original media is not used as fallback on videos, since it does not have the video tag. But I don't think the fallback is too important anyway.

PR #407, opened 2023-12-10
✨ Stop worker gracefully on sigterm

phlmn:


anuejn:
Overmind decides to kill the worker in the dev setup after a couple of seconds but I think this is okay

worker/run.py
```diff
@@ -49,56 +51,100 @@ def main():
...
+def run_sync_in_process(stop_watching_event: asyncio.Event, args):
+    p = Process(target=run_sync, args=(args,))
     p.start()
-    return p, event
+
+    def handle_signal(sig, *args):
+        if p.pid is None:
+            raise Exception("Process has no PID")
+
+        os.kill(p.pid, sig)
```
anuejn, line 82: what does this do?

worker/run.py
```diff
@@ -49,56 +51,100 @@ def main():
...
+
+        if sig == signal.SIGTERM:
+            stop_watching_event.set()
+        elif sig == signal.SIGUSR1:
+            stop_watching_event.clear()
+
+    signal.signal(signal.SIGTERM, handle_signal)
```
anuejn, line 89: why do we do it here again (compared to below?)

anuejn (commented):
Is it still possible to end the worker with a keyboard combination after this?
(e.g. ctrl-c ctrl-c)

worker/run.py
phlmn: it passes the signal from the reload watcher process to the actual worker process

worker/run.py
phlmn: this is the handling for the reload watcher process, which also is based on pythons multiprocess and therefor differs from the signal handling of the asyncio eventloop of the actual worker process below. The signal handling was a bit tricky due to the watcher process

phlmn:
> Overmind decides to kill the worker in the dev setup after a couple of seconds but I think this is okay

Yes this is intentional

worker/run.py
phlmn: (this also stops the file watching, otherwise it would continue to watch for file changes even after the actual worker process is stopped)

worker/run.py
anuejn: can I have comments explaining this?

worker/run.py
anuejn: can I have comments explaining this?

anuejn (approved):
Other than the comments LGTM

worker/run.py
phlmn: done

worker/run.py
phlmn: done

pajowu (approved):

PR #417, opened 2023-12-18
🐛 Fix and simplify reencode progress reporting

phlmn:


worker/transcribee_worker/reencode.py
```diff
@@ -46,31 +45,18 @@ def work(_):
...
-                stderr_data.append(raw_line.decode())
-                continue
-            if b"=" not in raw_line:
-                continue
-            key, value = raw_line.decode().strip().split("=", maxsplit=1)
-            progress_dict[key] = value.strip()
+        stderr_data = []
+        for raw_line in cmd.stderr:
+            line = raw_line.decode().rstrip()
+            logging.info(line)
```
pajowu, line 57: Why logging? This way we never see it in the logs sent to the db and can only investigate issues if we have access to the worker logs, which is not guaranteed

phlmn: Hm? stderr is still passed to the extra_data. The logging is mainly for local development.

pajowu (approved):

PR #435, opened 2024-04-27
🐛 Allow missing deps on loading initial document

phlmn:
Fixes loading some documents.


anuejn (approved):
I find it okay for the time being. Maybe we can find a better solution later

PR #439, opened 2024-05-03
💄 Sticky TopBar

### LINKED ISSUE #359: Feature Request: Sticky top navigation bar
In longer documents, the top nav bar scrolls out of view, whereas I would expect it / like it to be fixed and always visible.

phlmn:
closes #359

### DIFF
--- a/frontend/src/common/top_bar.tsx
+++ b/frontend/src/common/top_bar.tsx
@@ -1,4 +1,6 @@
+import clsx from 'clsx';
 import { useLocation } from 'wouter';
+
 import { storeAuthToken } from '../api';
 import { logout } from '../api/user';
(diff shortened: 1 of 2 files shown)


anuejn:
This looks broken. 
Screenshot 2024-05-17 at 14 31 36

Also I would prefer to have the header go out of the way while scrolling down and get into the frame when scrolling up. This keeps the application useable on high zoom, which is nice for accessebility.

phlmn:
@anuejn I cannot reproduce this... Did you do something special? 

hmm... I am not a huge fan of hiding the top bar when scrolling up and I really think there is enough space even on very high zoom.

anuejn (approved):

PR #440, opened 2024-05-03
📈 add active users metric

anuejn:


anuejn:
We decided against having that feature due to data-protecton concerns.

PR #449, opened 2024-12-03
 ✨ add option for timestamps in plaintext export

anuejn:


rroohhh (approved):

anuejn:
yeah Im fine with bikeshedding this another time. I think its just a matter of taste and the user who needs this wanted it like this :)

PR #452, opened 2025-01-07
📦️ Switch to uv and npm to build pure nix derviations

phlmn:


frontend/src/pages/about.tsx
```diff
@@ -18,7 +18,7 @@ export function AboutPage() {
           transcribee was founded from March until August 2023
           <img
             className="w-full"
-            src={pf_funding_svg}
+            // src={pf_funding_svg}
```
pajowu, line 21: Why that?

phlmn: Just WIP :) This needs to be moved/copied since we cannot import stuff from outside the frontend folder that easily when building the nix derivation

phlmn: fixed :)

rroohhh (approved):

PR #456, opened 2025-01-14
👷 Implement changes for flake based deployment

phlmn:


flake.nix
```diff
@@ -35,10 +35,38 @@
...
+
+            versionInfo = {
+              commitHash = if (self ? rev) then self.rev else self.dirtyRev;
+              commitDate = lib.readFile "${prev.runCommand "timestamp" { env.when = self.lastModified; } "echo -n `date -d @$when --iso-8601=s` > $out"}";
+            };
+          };
+        });
+
+      nixosModules.default = {
+        nixpkgs.overlays = [ self.overlays.default ];
```
rroohhh, line 65: Can we somehow get rid of the duplication in here with the `packages` below?

rroohhh (changes requested):
Why did you change the way the scripts / migrations were done?

flake.nix
phlmn: found a way :)

rroohhh (approved):

PR #490, opened 2025-11-26
worker: factor out video codec to be easily configurable

rroohhh:
Then we can switch (more easily) to `h264_videotoolbox` on the mac worker.


phlmn:
> Then we can switch (more easily) to `h264_videotoolbox` on the mac worker.

Is this a good idea? Encoding is already pretty fast and `h264_videotoolbox` is a bit crappy (e.g. it does not support `crf`)...

PR #499, opened 2025-12-28
worker: allow broken utf-8 in ffmpeg output

rroohhh:
current VOC pipeline for 39c3 cuts long description fields without a regard for utf-8 encoding, so the embedded description will not be valid utf-8


anuejn (approved):
very nice :)

PR #510, opened 2026-01-07
🐣 add tools for reflowing the transcript into one paragraph per sentence / speaker

anuejn:


frontend/src/editor/automerge_websocket_editor.ts
```diff
@@ -54,7 +58,9 @@ export function useAutomergeWebsocketEditor(
     const createNewEditor = (doc: Automerge.Doc<Document>) => {
       const baseEditor = createEditor();
       const editorWithReact = withReact(baseEditor);
-      const editor = withHistory(withAutomergeDoc(editorWithReact, Automerge.init()));
+      const editor = withHistory(
+        withAutomergeDoc(editorWithReact, Automerge.init()),
+      ) as EditorWithWebsocket;
```
rroohhh: This feels a bit hack, I feel like the "clean" way would be a `withWebsocket`, but as we want to drop this code soon anyways (by replacing with `loro`) I would be fine merging like this.

frontend/src/editor/text_tools.tsx
```diff
@@ -0,0 +1,182 @@
+import { TbHammer } from 'react-icons/tb';
```
rroohhh: Can we have tests for these? :)

frontend/src/editor/text_tools.tsx
```diff
@@ -0,0 +1,182 @@
...
+const punctuations = ['.', '?', '!'];
+const non_punctuations = ['...'];
+function containsSentenceEnd(text: string) {
+  return (
+    punctuations.some((punct) => text.includes(punct)) &&
+    !non_punctuations.some((np) => text.includes(np))
+  );
+}
+
+export function TextTools({ editor }: { editor: EditorWithWebsocket }) {
```
rroohhh: Should we add a warning to these if applied to a document that is in a non latin style language?

frontend/src/editor/text_tools.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    };
+                  }
+                });
+                if (currentParagraph.children.length > 0) {
+                  newChildren.push(currentParagraph);
+                }
+              }
+            };
+            doc.children.forEach((paragraph) => {
+              let minPauseBetweenSentences = initial; // this gets reduced with every additional token
```
rroohhh: Why does it get reduces with every additional token?

frontend/src/editor/text_tools.tsx
```diff
@@ -0,0 +1,182 @@
...
+                    ...paragraph,
+                    children: [],
+                  };
+                }
+              });
+              if (currentParagraph.children.length > 0) {
+                addNewChild(currentParagraph);
+              }
+            });
+            doc.children = newChildren;
```
rroohhh: I think doing it this way totally fucks up collaborative editing...

PR #512, opened 2026-01-08
Fixes for webVTT export

phlmn:


worker/transcribee_worker/webvtt/export_webvtt.py
```diff
@@ -252,7 +268,6 @@ def par_len(pars):
             fits = par_len(pars) + this_par_len < character_limit_pack
             if fits:
                 pars.append(par)
-                last_speaker = speaker
```
rroohhh: oops, good catch

worker/transcribee_worker/webvtt/export_webvtt.py
```diff
@@ -148,7 +148,21 @@ def split_atoms(atoms):
...
+    if not par.children:
+        return par
+
+    if par.children[-1].text.endswith(" "):
+        return par
+
+    children = [child for child in par.children]
+    children[-1].text += " "
+
+    return par.model_copy(update={"children": children})
```
rroohhh: Why, you can just modify inplace, no?

phlmn: Just not a fan of side effects

PR #527, opened 2026-05-16
Build backend with docker and fix dev container

phlmn:


.github/workflows/publish-docker.yml
```diff
@@ -1,4 +1,4 @@
-name: Create and publish a Docker image
+name: Publish docker images
```
anuejn, line 1: build &

packaging/setup_backend.sh
```diff
@@ -7,5 +7,5 @@ if [[ -z "TRANSCRIBEE_BACKEND_DATABASE_URL" ]]; then
   wait4x postgresql "postgres://@/transcribee?host=$(pwd)/../backend/db/sockets/"
 fi
 poe -C ../backend/ migrate
-poe -C ../backend/ admin create_user --user test --pass test
+poe -C ../backend/ admin create_user --user admin --pass admin
```
anuejn: y tho?

compose-dev.yml
```diff
@@ -0,0 +1,29 @@
...
+      POSTGRES_PASSWORD: transcribee
+      POSTGRES_USER: transcribee
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -d transcribee -U transcribee"]
+      interval: 1s
+      timeout: 5s
+      retries: 10
+    volumes:
+      - postgres-data:/var/lib/postgresql
+
```
anuejn: y no redis here?

anuejn (commented):
could you also delete the nix folder?

anuejn (approved):

.python-version
anuejn, line 1: I would like to not have this as a toplevel file. I would be okay to leave the versions separate (we already have to change it in multiple places anyway) and this suggests, that this is the only place to change it (it is not).

doc/development_setup.md
anuejn, line 1: I actually like the dev.sh being in packaging

docker-devshell.sh
anuejn, line 1: could you move this to packaging/ and document this somewhere?

anuejn (changes requested):

docker-devshell.sh
phlmn: deleted in favor of dev container

.python-version
phlmn: The other `.python-version` files are symlinked to this. While we still need to change the version in other places (e.g. flake.nix) uv should always verify that we use the right python version based on this single file.

doc/development_setup.md
phlmn: ok

.python-version
anuejn: ok

anuejn (approved):
cool, looking forward to using this!

PR #547, opened 2026-06-18
Fix local deployment

faulesocke:
Local deployment failed because the worker token can not be written in run.sh:14 since the volume mounted in compose-local.yml is owned by root. Changing ownership of the volume directory here mounts the volume with correct permissions and allows creating the worker token file at runtime.

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -49,7 +49,9 @@ COPY --from=frontend-builder /app/frontend/dist ./public
 COPY packaging/docker/run.sh /run.sh
 
 ENV STORAGE_PATH=/var/transcribee/storage
+ENV TOKEN_PATH=/var/transcribee/worker_token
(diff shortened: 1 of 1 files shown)


anuejn:
Hey, thank you for trying transcribee and fixing this :)
Sadly I cannot really reproduce the issue you describe here :(
Running `docker compose -f packaging/compose-local.yml up` from the root of the repository works on my machine. Could you describe what I need to do to reproduce the issue you experience?

anuejn:
maybe it has something to do with the fact that I am using wired mac docker?
what system are you running on?

faulesocke:
For reproduction, maybe try removing the volume, then try again. I tried this on 2 Linux boxes, one with Ubuntu and one with Void, both show the same permission problem.

phlmn:
I can reproduce this on my system after pruning docker. We should think about applying permission fixes in the entry script and then switch the user with something like `runuser` or `gosu`. But I think this is a practical fix for now. Thank you!

phlmn (approved):

PR #550, opened 2026-06-25
worker: Add opt-in TwelveLabs Pegasus transcription backend

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## What this adds

An **opt-in** transcription backend that delegates transcription to the [TwelveLabs](https://twelvelabs.io) **Pegasus** video-understanding model, as an alternative to the default local whisper backend.

- New module `worker/transcribee_worker/pegasus_transcribe.py`: indexes the document media into a TwelveLabs index, asks Pegasus for a timed transcript, and parses the result into transcribee `Paragraph`s. It mirrors the existing `whisper_transcribe` async/`SubmissionQueue` pattern.
- New settings in `config.py`: `TRANSCRIPTION_BACKEND` (`whisper` default / `pegasus`), `TWELVELABS_API_KEY`, `PEGASUS_MODEL`, `PEGASUS_INDEX_NAME`, `PEGASUS_MAX_TOKENS`.
- `worker.transcribe()` branches on the backend; the whisper path is behaviourally unchanged.
- The `twelvelabs` SDK is an **optional** dependency (the `pegasus` extra), so the default install and whisper backend are untouched.

## Why it helps

It gives self-hosters a fully server-side transcription option that needs no local model download or GPU — useful for lightweight deployments. It also slots TwelveLabs' multimodal video understanding into transcribee's existing worker architecture without disrupting anything.

## Opt-in / non-breaking

Default behaviour is identical: `TRANSCRIPTION_BACKEND` defaults to `whisper`, the SDK is only pulled in via `uv sync --extra pegasus`, and the Pegasus code path is only reached when explicitly enabled.

## Implementation note

Pegasus is a *video* model and the platform requires indexed media to have a video track of at least 360x360. transcribee's common case is audio, so audio-only media is wrapped in a minimal black video track before indexing. The repo moved transcoding to PyAV (commit d7f965a), but synthesizing a fresh video track from scratch in PyAV is materially more code/fragile than a single `ffmpeg` invocation, and `ffmpeg` is already a bundled native dependency — so this one wrapper step shells out to `ffmpeg`. Happy to convert it to PyAV if you'd prefer to keep zero ffmpeg-CLI usage.

## How it was tested

- `worker/tests/test_pegasus_transcribe.py`: unit tests for the JSON→`Paragraph` parser (fence stripping, offset application, duration clamping, empty-segment skipping, malformed-input rejection) — no network.
- A live end-to-end test gated on `TWELVELABS_API_KEY` (skips without it) that runs the full upload → index → analyze → parse flow on the bundled `tests/data/sample.mp3` and asserts a non-empty timed paragraph. Verified passing against the real API.
- `ruff`, `black`, and `pyright` pass on the changed worker files.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/worker/README.md
+++ b/worker/README.md
@@ -26,3 +26,27 @@ The most important settings are explained here:
 The worker needs multiple models, which will automatically be downloaded if they to not exist yet.
 They will by default be stored in `./.data/models`.
 You can change this directory by setting `MODELS_DIR`
+
+### Transcription backends
+
+By default transcription runs locally with whisper (`TRANSCRIPTION_BACKEND=whisper`).
+
+As an opt-in alternative you can delegate transcription to the
+[TwelveLabs](https://twelvelabs.io) Pegasus video-understanding model. This runs
+server-side and requires no local model download. To enable it:
+
+```shell
+uv sync --extra pegasus            # installs the twelvelabs SDK
+export TRANSCRIPTION_BACKEND=pegasus
+export TWELVELABS_API_KEY=<your-key>   # free tier at https://twelvelabs.io
+```
+
+Relevant settings (see `config.py`):
+
+- `TRANSCRIPTION_BACKEND` &ndash; `whisper` (default) or `pegasus`
+- `TWELVELABS_API_KEY` &ndash; required when using the `pegasus` backend
+- `PEGASUS_MODEL` &ndash; Pegasus model name (default `pegasus1.2`)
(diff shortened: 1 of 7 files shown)


rroohhh:
This is not something we are interested in.

PR #558, opened 2026-07-06
fix: add support for video with non square pixels

anuejn:


rroohhh (commented):
This means non square "pixels", right?

rroohhh (commented):
Is this the correct way to handle this, vs setting the sar on the output?

anuejn:
Yes, I like the idea of not having "weired" video after the first re-encoding

rroohhh (approved):
Okay :)

Issue #518, opened 2026-02-08
Transcription failes when there is no audio


rroohhh:
```
INFO:faster_whisper:Processing audio with duration 00:25.963
INFO:faster_whisper:VAD filter removed 00:25.963 of audio
INFO:faster_whisper:Detected language 'en' with probability 0.52
WARNING:root:Worker failed with exception
Traceback (most recent call last):
  File "/nix/store/rmz3azc2p4vh87is42qfjkzimmi1nj48-transcribee-worker-env/lib/python3.11/site-packages/transcribee_worker/whisper_transcribe.py", line 34, in move_space_to_prev_tok
en
    last_paragraph = next(iter)
                     ^^^^^^^^^^
StopIteration

The above exception was the direct cause of the following exception:
```

PR #559, opened 2026-07-06
fix worker crashing on documents without speech

anuejn:


rroohhh:
fixes #518

rroohhh (approved):
Whatever it takes

PR #560, opened 2026-07-06
fix: improve heuristic for not combining paragraphs

anuejn:


rroohhh (commented):
Can we get a test for this? 
Also can you describe what you actually changed?

anuejn:
huh, it seems like this is actually also not yet working. A test document (https://media.ccc.de/v/39c3-denkangebot-rainer-mhlhoff-ber-ki-und-autoritre-sehnschte-im-silicon-valley#t=1811) now transcribes in a way that is not better and hangs the editor quite badly. Ill have to investigate some more

anuejn:
should be good to go now :)

rroohhh (approved):
Do we remember, why we had the "until the end of the document" behavior previously?

anuejn:
The assumption was, that the whisper model will not get out of its corner of not producing sentence breaks anymore. This is mostly true but our solution lead to some even worse failure when the heuristic triggered accidentally. Basically we made a less-than-optimal trade-off.

anuejn:
this is a good test document where this behaviour is displayed: https://media.ccc.de/v/38c3-breaking-the-mirror-a-look-at-apple-s-new-iphone-remote-control-feature

PR #562, opened 2026-07-08
editor: remove start, end, conf and conf_ts props on paste

rroohhh:
This does not really work well at the moment, because we assume all nodes to have a `start` and `end`, so export is gated on it and the timeline sidebar displays "unknown" as timestamp.

To make this proper, we should probably insert the `start` and `end`, so that we fit inbetween our neigbours, but this is a bit tricky to do at this point....


rroohhh:
It also breaks the worker side export. 

We currently specify, that `start` and `end` are allowed to be null in our document format, so this is technically just our frontend and our backend code not adhering to the document format, so we could also fix those relying on `start` / `end` being always set, but not sure that is the more sensible path, because looking at previous / following nodes in slate sucks.

PR #570, opened 2026-07-17
desktop: Add linting and formatting checks

phlmn:


.github/workflows/lint.yml
```diff
@@ -20,8 +20,11 @@ jobs:
       - name: install pre-commit
         run: pip install pre-commit ruff poethepoet
 
-      - name: install our dependencies
+      - name: install web dependencies
         run: ./packaging/install_dependencies.sh
```
anuejn, line 25: do we need to install the ui-common dependencies here as well?

phlmn: They get installed in install_dependencies.sh already

anuejn (approved):

PR #580, opened 2026-07-23
pre-commit: factor out package-manager run checks

anuejn:
common checks that are run through the package manager are now invoked through a bespoke python script. this allows us to execute them in paralell (making it faster), installing the dependencies before running the checks (making it more correct) and allowing us to automatically discover projects (preventing stupid mistakes we had in the past)

### DIFF
--- a/desktop/src-tauri/build.rs
+++ b/desktop/src-tauri/build.rs
@@ -1,3 +1,22 @@
+use std::{
+    fs::{create_dir_all, remove_dir_all},
+    process::Command,
+};
+
+use tauri_build::is_dev;
+
 fn main() {
+    // in dev mode we simply start the worker from ../../worker with uv while in production
+    // we use the bundled worker. We still need to create an empty worker/ dir in dev to make
+    // the tauri build process happy.
+    if is_dev() {
+        let _ = remove_dir_all("worker");
+        create_dir_all("worker").expect("could not create worker/ dir");
+    } else {
+        Command::new("python")
+            .args(["src-tauri/prepare_worker.py"])
+            .output()
+            .expect("failed to build worker bundle");
+    }
     tauri_build::build()
 }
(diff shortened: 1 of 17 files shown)


desktop/src-tauri/build.rs
```diff
@@ -1,3 +1,19 @@
+use std::{
+    fs::{create_dir_all, remove_dir_all},
+    process::Command,
+};
+
+use tauri_build::is_dev;
+
 fn main() {
+    if is_dev() {
```
phlmn, line 12: this needs some comments. I can only guess why we need an empty worker folder.

desktop/src-tauri/build.rs
```diff
@@ -1,3 +1,19 @@
...
+        let _ = remove_dir_all("worker");
+        create_dir_all("worker").expect("could not create worker/ dir");
+    } else {
+        Command::new("python")
+            .args(["src-tauri/prepare_worker.py"])
+            .output()
+            .expect("failed to build worker bundle");
+    }
```
phlmn, line 20: Can we set a custom rust cfg value here which we then use in `worker_plugin.rs`?

phlmn (changes requested):

phlmn:
I do wonder if we really want to run tests in the pre-commit hooks

anuejn:
Yeah, I have no strong opinions on this. I would like to run the tests in the ci though :).
On my machine it does not make a big difference time-wise to include them or not (~12.6s vs ~8.9s for a full run of pre-commit-projects.py with warm cache).

phlmn:
> Yeah, I have no strong opinions on this. I would like to run the tests in the ci though :). On my machine it does not make a big difference time-wise to include them or not (~12.6s vs ~8.9s for a full run of pre-commit-projects.py with warm cache).

Let's try it this way and change it if it annoys us :)

phlmn (approved):

PR #603, opened 2026-08-05
desktop: add support for reencoding audio & speaker identification

anuejn:
this fixes using transcribee desktop with media files that cannot be natively opened by the browser

also this pr enables spearker identification on desktop


worker/transcribee_worker/worker.py
```diff
@@ -293,6 +303,11 @@ def work(_):
                 None, self.add_document_media_file, task, output_path, tags
             )
 
+            if settings.WORKER_TYPE == "desktop":
+                # we only reencode one version for desktop (no need to save bandwidth)
+                # in only loading the audio if video preview is disabled
+                break
+
```
phlmn: not a fan of selecting the profile like this

desktop/src/pages/home.tsx
```diff
@@ -227,6 +241,35 @@ function ProgressPie({
...
+    task_type: 'IDENTIFY_SPEAKERS' | 'TRANSCRIBE' | 'REENCODE' | 'EXPORT';
+    state: 'NEW' | 'ASSIGNED' | 'COMPLETED' | 'FAILED';
+    current_attempt: { progress: number | null } | null;
+  }[],
+) {
+  const weights = {
+    IDENTIFY_SPEAKERS: 0.1,
+    TRANSCRIBE: 2.0,
+    REENCODE: 0.1,
+    EXPORT: 0.01,
```
phlmn: EXPORT should not affect transcription progress

desktop/src-tauri/src/file_handling.rs
```diff
@@ -309,29 +300,44 @@ pub fn forget_document(
...
+    let Some(media_file) = document
+        .media_files
+        .iter()
+        .find(|m| m.tags.iter().any(|t| t == "browser_compatible"))
+    else {
+        let focused_window =
+            focused_window(&app_handle).ok_or(anyhow!("could not get focused window"))?;
+        focused_window
+            .dialog()
+            .message("Could not save because no suitable media file was found. Please try again in a few seconds when transcribee has prepared a suitable file.")
```
phlmn: Maybe something like "Could not save, because the media file is still being processed. [...]" is easier to understand.

desktop/src-tauri/src/file_handling.rs
```diff
@@ -309,29 +300,44 @@ pub fn forget_document(
...
+            .kind(MessageDialogKind::Error)
+            .title("could not save")
```
phlmn: title case

desktop/src-tauri/src/lib.rs
```diff
@@ -197,6 +198,25 @@ pub fn run() {
                     )
             });
 
+            // forget all documents that have unfinished transcription jobs
+            let app_handle = app.app_handle().clone();
+            app.update_documents(move |mut documents| {
+                let app_handle = app_handle.clone();
+                let worker_adapter = worker_adapter.clone();
+                documents.retain(move |doc| {
+                    let keep = doc.worker_tasks.iter().all(|task| task.state == Completed);
```
phlmn: I would argue this is a footgun. I would prefer this to only remove documents that are pretty much useless. Not documents one has been working on and any task that happened after that failed. Think of export additional reencoding, speaker detection and something like that.

phlmn (changes requested):

desktop/src-tauri/src/lib.rs
phlmn: Or mark non-completed tasks as failed when starting the app and add UI for broken documents instead of deleting them. So not all progress is lost when transcribee should crash on long documents in the last second

anuejn: true

desktop/src/pages/home.tsx
```diff
@@ -78,7 +92,24 @@ function TranscriptionQueue({ documents }: { documents: Document[] }) {
...
+                fallbackPlacements={['bottom', 'top']}
+                tooltipText={
+                  <span>
+                    Automatic transcription did not complete! This may happen when transcribee is
+                    closed while a task is running.
+                  </span>
+                }
+                className="ml-2"
+              >
+                <IoIosCloseCircle className="text-red-500 shrink-0" size={21} />
```
phlmn: can i have text-red-600?

desktop/src-tauri/src/lib.rs
```diff
@@ -143,22 +143,74 @@ pub fn run() {
...
+                                .unwrap();
+                        },
+                    )
             });
 
+            // mark all jobs that are unfinished as aborted
+            app.update_documents(move |mut documents| {
+                for doc in &mut documents {
+                    for task in &mut doc.worker_tasks {
+                        if task.state != Completed {
```
phlmn: This should only affect New and Assigned state

phlmn (changes requested):

phlmn (approved):

PR #614, opened 2026-08-11
Workaround linux custom protocol issue

anuejn:
also refactor the desktop rust code to be a bit more organized

### DIFF
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -1,18 +1,10 @@
 use crate::cmd::install_cmds;
-use crate::file_handling::{DocumentsStoreExt, MediaFile};
-use crate::media_file_serve::install_media_file_serve;
 use crate::window::create_or_focus_main_window;
 use colored::Color;
-use log::{warn, Level};
-use serde_json::json;
-use std::time::Duration;
+use log::Level;
 use tauri::RunEvent;
-use tauri::{Emitter, Manager};
 use tauri_plugin_log::fern;
-use tauri_plugin_store::StoreExt;
 use tauri_plugin_window_state::StateFlags;
-use worker_adapter::state::TaskState::{Aborted, Assigned, New};
-use worker_adapter::WorkerAdapter;
 
 mod cmd;
 mod cmd_error;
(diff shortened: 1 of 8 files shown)


desktop/src-tauri/src/lib.rs
```diff
@@ -65,78 +58,20 @@ pub fn run() {
...
         .plugin(tauri_plugin_opener::init())
-        .plugin(tauri_plugin_store::Builder::default().build())
-        .plugin(worker_plugin::init());
+        .plugin(tauri_plugin_store::Builder::default().build());
+
     let builder = install_cmds(builder);
+    let builder = builder.plugin(menu::init());
+    let builder = builder.plugin(file_handling::init());
+    let builder = builder.plugin(media_file_serve::init());
+    let builder = builder.plugin(worker_plugin::init());
```
phlmn: I think this would be nicer if chained like above

desktop/src-tauri/src/media_file_serve.rs
```diff
@@ -1,39 +1,105 @@
...
+                            .body(e.to_string().as_bytes().to_vec())
+                            .unwrap(),
+                    ),
+                }
+            })
+            .setup(|app, _| {
+                app.manage(MediaFileBase("media://localhost".to_string()));
+                Ok(())
+            })
+    };
```
phlmn: I think we should just use the http endpoint for all platforms to reduce complexity

desktop/src-tauri/src/media_file_serve.rs
```diff
@@ -1,39 +1,105 @@
...
+                    listener.local_addr().unwrap()
+                );
+
+                app.manage(MediaFileBase(format!(
+                    "http://{}",
+                    listener.local_addr().unwrap()
+                )));
+
+                #[axum::debug_handler]
+                async fn get_response(State(app): State<AppHandle>, request: Request) -> Response {
```
phlmn, line 50: add simple mechanism to secure this endpoint beyond needing to know the document uuid

phlmn (approved):

desktop/src-tauri/src/media_file_serve.rs
rroohhh, line 46: Why can this not be "localhost"?

rroohhh: (And if it can be, `MediaFileBase` could be removed again)

anuejn: it needs to know the port (which is randomly choosen)

PR #616, opened 2026-08-12
chore: add nix mesa on non-nixos handling to dev shell

rroohhh:
Using nix mesa on non-nixos needs additional environment variables, which nixGL sets up.


anuejn (approved):
nice :)

PR #621, opened 2026-08-13
little fixes

anuejn:


desktop/package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.1.0",
   "scripts": {
+    "prepare": "npm --prefix ../ui-common/ ci",
```
phlmn, line 6: Does this work without internet?

phlmn: ...or should we use install? Not sure about this, but let's try

phlmn (approved):

desktop/package.json
anuejn: I thought `ci` is the better choice here because it does not change the package.lock which would be unexpected. I am not sure if it needs to work without internet as the install that triggers this prepare hook also does not work without internet?

PR #643, opened 2026-08-21
Macos lint

rroohhh:


rroohhh:
Doesn't work because the transcribee web backend tests are linux only :(

